PR #47, opened 2016-09-12
fix: Use provider HC when Eureka enabled

e4r7hbug:
Resolves a FIXME that has needed to be done for now. Now that
scaffolding is in place, use that to set the option.

See also: PSOBAT-2465

### DIFF
--- a/src/foremast/pipeline/construct_pipeline_block.py
+++ b/src/foremast/pipeline/construct_pipeline_block.py
@@ -14,6 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 """Construct a block section of Stages in a Spinnaker Pipeline."""
+import collections
 import copy
 import json
 import logging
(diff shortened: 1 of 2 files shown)


ndcampbell:
Just for clarification, what is the intended behavior from this change?

e4r7hbug:
This fix will check the `Consider only Amazon health` box in the Deployment Stage by default. This way, Spinnaker will look at the ASG Health Check set to EC2. Currently, we need to specify `{"provider_healthcheck": {"Amazon": true}}` separately.

e4r7hbug:
Wait, Sijis pointed out another issue. Will need to create a provider in a dictionary.

ndcampbell:
Ah, what will the provider in a dictionary do?

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -25,6 +25,32 @@
...
+
+    for provider, active in providers.items():
+        if active:
+            provider_healthcheck.append(provider.capitalize())
+    LOG.info('Provider healthchecks: {0}'.format(provider_healthcheck))
+
+    if len(provider_healthcheck) > 0:
+        has_provider_healthcheck = True
+
+    return provider_healthcheck, has_provider_healthcheck
```
sijis: maybe we should use `namedtuples`?
``` python
ProviderHealthcheck = namedtuple('ProviderHealthcheck', ['provider', 'healthcheck'])
return ProviderHealthcheck(provider_healthcheck, has_providerhealthcheck)
```

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -83,22 +109,14 @@ def construct_pipeline_block(env='',
...
-    provider_healthcheck = []
-    for provider, active in settings['asg']['provider_healthcheck'].items():
-        if active:
-            provider_healthcheck.append(provider.capitalize())
-    LOG.info('Provider healthchecks: {0}'.format(provider_healthcheck))
-
-    has_provider_healthcheck = False
-    if len(provider_healthcheck) > 0:
-        has_provider_healthcheck = True
+    provider_healthcheck, has_provider_healthcheck = check_provider_healthcheck(settings)
```
sijis: Assuming we go with `namedtuple`, we have to adjust this to something like:

``` python
provider_healthcheck = check_provider_healthcheck(settings)
```

sijis:
:+1:

ndcampbell:
LGTM 👍

e4r7hbug:
@ndcampbell: The Provider dictionary was stubbing for future. We only have Amazon right now, but we don't know what the checkbox name will be for Azure or Google.

PR #48, opened 2016-09-12
fix: Fallback `git describe` string for tox

e4r7hbug:
When `tox` runs, `git describe` fails to generate the version number.
This provides a safe default in the case when the command fails to
generate a version number.


sijis:
:+1:

PR #49, opened 2016-09-12
Fix test cases as default path of api gateway resource changed

sijis:
This fixes the test case since at some point the default path for a resource name changed.


ndcampbell:
👍

e4r7hbug:
Getting this merged in.

Issue #50, opened 2016-09-13
Lambda create_function() should retry


e4r7hbug:
Due to the delay that happens when modifying a Role's Inline Policy, we need to retry a few times.

PR #53, opened 2016-09-13
fix: Add retry to Lambda creation

e4r7hbug:
Need to compensate for lag in IAM Role Policy changes.

See also: #50

### DIFF
--- a/src/foremast/awslambda/awslambda.py
+++ b/src/foremast/awslambda/awslambda.py
@@ -119,6 +119,7 @@ class LambdaFunction(object):
 
         LOG.info("Successfully updated Lambda function")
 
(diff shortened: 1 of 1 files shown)


ndcampbell:
This change looks good to me and follows our "retry" patterns, but I am kind of afraid of it falling apart at some point. Wouldnt it be better to query for the IAM Role Policy and make sure the change took effect instead of just retrying? 

I think it can be merged in, just wanted to voice my concerns (and I know it isn't the first time 😛 )

sijis:
@ndcampbell I understand the concern around retry and its valid. I will say, that the good thing about it, is that if AWS (or Spinnaker) is having API issues, this allows us to safely retry the action until it completes.

Besides that, I think its a safe enough change to :+1: .

PR #54, opened 2016-09-13
Fix/lambda type plus service

e4r7hbug:
Automatically add lambda service for IAM Policy.


sijis:
This makes sense, as it simplifies the user's need to also specify lambda in the `services` key.
:+1:

ndcampbell:
👍  LGTM

Issue #55, opened 2016-09-14
Lambda creation/code update should create ALIAS with ENV name


asibin:
When creating Lambda function or updating lambda function we should alias $latest with ENV name so developers can easily fetch environment name and apply proper configs. That way they do not need to know account numbers or any other info.

Having lambda configs for all environments locally shortens the execution time (by not having to lookup external configs/services) and enables us to have single artifact for multiple ENVs for lambdas that need to have multiple different configs (for example ones that query Eureka for services). 

This in turn needs to apply triggers to ENV Alias and not to $latest as it was the case which will leave current triggers defined on $latest for already deployed functions. 

I can submit PR for this functionality if everyone agrees we need this kind of functionality.

asibin:
Lambda aliasing is tested and working as expected. Closing issue.

PR #60, opened 2016-09-20
Feat/lambda aliasing

asibin:
Fixes issue #55.

`get_lambda_arn()` is not used anymore (new function is `get_lambda_alias_arn`)  but i haven't deleted it because it maybe could be useful to some other use case. If you want me to delete it i will.

### DIFF
--- a/src/foremast/awslambda/awslambda.py
+++ b/src/foremast/awslambda/awslambda.py
@@ -62,7 +62,8 @@ class LambdaFunction(object):
         return exists
 
     def _check_lambda_alias(self):
-        """ Checks if lambda alias exists
+        """Check if lambda alias exists.
+
         Returns:
             True if alias exists
             False if alias does not exist
@@ -107,12 +108,12 @@ class LambdaFunction(object):
             sg_ids.append(sg_id)
         return sg_ids
 
-    @retries(max_attempts=3, wait=1)
+    @retries(max_attempts=3, wait=1, exceptions=(boto3.exceptions.botocore.exceptions.ClientError))
     def create_alias(self):
-        """Create lambda alias with env name and points it to $LATEST"""
+        """Create lambda alias with env name and points it to $LATEST."""
+        LOG.info('Creating alias %s', self.env)
 
         try:
-            LOG.info('Creating alias %s', self.env)
             self.lambda_client.create_alias(
                 FunctionName=self.app_name,
                 Name=self.env,
@@ -120,23 +121,23 @@ class LambdaFunction(object):
                 Description='Alias for {}'.format(self.env)
             )
         except boto3.exceptions.botocore.exceptions.ClientError as error:
-            LOG.debug(str(error))
+            LOG.debug('Create alias error: %s', error)
             LOG.info("Alias creation failed. Retrying...")
             raise
 
-    @retries(max_attempts=3, wait=1)
+    @retries(max_attempts=3, wait=1, exceptions=(boto3.exceptions.botocore.exceptions.ClientError))
     def update_alias(self):
-        """Updates lambda alias to point to $LATEST"""
+        """Update lambda alias to point to $LATEST."""
+        LOG.info('Updating alias to point to $LATEST', self.env)
 
         try:
-            LOG.info('Updating alias to point to $LATEST', self.env)
             self.lambda_client.update_alias(
                 FunctionName=self.app_name,
                 Name=self.env,
                 FunctionVersion='$LATEST'
             )
         except boto3.exceptions.botocore.exceptions.ClientError as error:
-            LOG.debug(str(error))
+            LOG.debug('Update alias error: %s', error)
             LOG.info("Alias update failed. Retrying...")
             raise
 
@@ -148,9 +149,9 @@ class LambdaFunction(object):
             vpc_config (dict): Dictionary of SubnetIds and SecurityGroupsIds for using
                                a VPC in lambda
         """
-        try:
-            LOG.info('Updating code for lambda function: %s', self.app_name)
+        LOG.info('Updating code for lambda function: %s', self.app_name)
 
+        try:
             self.lambda_client.update_function_configuration(
                 FunctionName=self.app_name,
                 Runtime=self.runtime,
@@ -190,9 +191,9 @@ class LambdaFunction(object):
         with open('lambda-holder.zip', 'rb') as openfile:
             contents = openfile.read()
 
-        try:
-            LOG.info('Creating lambda function: %s', self.app_name)
+        LOG.info('Creating lambda function: %s', self.app_name)
 
+        try:
             self.lambda_client.create_function(
                 FunctionName=self.app_name,
                 Runtime=self.runtime,
@@ -219,13 +220,13 @@ class LambdaFunction(object):
     def create_lambda_function(self):
         """Create or update Lambda function."""
         vpc_config = self._vpc_config()
+
         if self._check_lambda():
             self.update_function_configuration(vpc_config)
-
-            if self._check_lambda_alias():
-                self.update_alias()
-            else:
-                self.create_alias()
         else:
             self.create_function(vpc_config)
+
+        if self._check_lambda_alias():
+            self.update_alias()
+        else:
             self.create_alias()
--- a/src/foremast/awslambda/awslambda.py
+++ b/src/foremast/awslambda/awslambda.py
@@ -1,3 +1,19 @@
+#   Foremast - Pipeline Tooling
+#
+#   Copyright 2016 Gogo, LLC
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
 import logging
 import zipfile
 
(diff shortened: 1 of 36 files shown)


src/foremast/utils/awslambda.py
```diff
@@ -33,6 +33,32 @@ def get_lambda_arn(app, account, region):
                                                                                                      region))
 
 
+def get_lambda_alias_arn(app, account, region):
```
sijis: I'm not seeing a test case around this method and its obviously used in several locations.

src/foremast/awslambda/awslambda.py
```diff
@@ -61,6 +61,22 @@ def _check_lambda(self):
             pass
         return exists
 
+    def _check_lambda_alias(self):
```
sijis: This should probably have a simple test case around it too.

src/foremast/awslambda/awslambda.py
```diff
@@ -140,6 +191,8 @@ def create_function(self, vpc_config):
             contents = openfile.read()
 
         try:
+            LOG.info('Creating lambda function: %s', self.app_name)
```
e4r7hbug: Move INFO outside of `try`.

src/foremast/awslambda/awslambda.py
```diff
@@ -161,12 +214,18 @@ def create_function(self, vpc_config):
...
-        LOG.info("Successfully created Lambda function")
+        LOG.info("Successfully created Lambda function and alias")
 
     def create_lambda_function(self):
         """Create or update Lambda function."""
         vpc_config = self._vpc_config()
         if self._check_lambda():
             self.update_function_configuration(vpc_config)
+
+            if self._check_lambda_alias():
```
e4r7hbug: Can we move this `if` lower to avoid nesting?
```python
if _check_lambda():
    update_function()
else:
    create_function()

if _check_lambda_alias():
    update_alias()
else:
    create_alias()
```
It might even make more sense to move the if statements into the `checks()`.
```python
def _check_lambda(self):
    self.vpc_config = self._vpc_config()

    try:
        self.lambda_client.get_function(FunctionName=self.app_name)
        self.update_function_configuration()
    except boto3.exceptions.botocore.exceptions.ClientError:
        self.create_function()

def _check_lambda_alias(self):
    if found:
        self.update_alias()
    else:
        self.create_alias()
```

src/foremast/awslambda/awslambda.py
```diff
@@ -100,6 +149,8 @@ def update_function_configuration(self, vpc_config):
                                a VPC in lambda
         """
         try:
+            LOG.info('Updating code for lambda function: %s', self.app_name)
```
e4r7hbug: Move INFO to outside of `try`.

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
...
+
+        try:
+            LOG.info('Updating alias to point to $LATEST', self.env)
+            self.lambda_client.update_alias(
+                FunctionName=self.app_name,
+                Name=self.env,
+                FunctionVersion='$LATEST'
+            )
+        except boto3.exceptions.botocore.exceptions.ClientError as error:
+            LOG.debug(str(error))
```
e4r7hbug: More info in DEBUG.
```python
LOG.debug('Update alias error: %s', error)
```

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
...
+            LOG.info("Alias creation failed. Retrying...")
+            raise
+
+    @retries(max_attempts=3, wait=1)
+    def update_alias(self):
+        """Updates lambda alias to point to $LATEST"""
+
+        try:
+            LOG.info('Updating alias to point to $LATEST', self.env)
```
e4r7hbug: Move `LOG.info()` to outside of `try`.

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
...
+                Name=self.env,
+                FunctionVersion='$LATEST',
+                Description='Alias for {}'.format(self.env)
+            )
+        except boto3.exceptions.botocore.exceptions.ClientError as error:
+            LOG.debug(str(error))
+            LOG.info("Alias creation failed. Retrying...")
+            raise
+
+    @retries(max_attempts=3, wait=1)
```
e4r7hbug: Include `exceptions` here too.

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
...
+        try:
+            LOG.info('Creating alias %s', self.env)
+            self.lambda_client.create_alias(
+                FunctionName=self.app_name,
+                Name=self.env,
+                FunctionVersion='$LATEST',
+                Description='Alias for {}'.format(self.env)
+            )
+        except boto3.exceptions.botocore.exceptions.ClientError as error:
+            LOG.debug(str(error))
```
e4r7hbug: Need more help to know what DEBUG is telling me.
```python
LOG.debug('Create alias error: %s', error)
```

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
             sg_ids.append(sg_id)
         return sg_ids
 
+    @retries(max_attempts=3, wait=1)
+    def create_alias(self):
+        """Create lambda alias with env name and points it to $LATEST"""
+
+        try:
+            LOG.info('Creating alias %s', self.env)
```
e4r7hbug: Move `LOG.info()` to outside of `try`.

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
             sg_ids.append(sg_id)
         return sg_ids
 
+    @retries(max_attempts=3, wait=1)
+    def create_alias(self):
+        """Create lambda alias with env name and points it to $LATEST"""
```
e4r7hbug: Docstring should end in period and no newline after.

src/foremast/awslambda/awslambda.py
```diff
@@ -91,6 +107,39 @@ def _get_sg_ids(self):
             sg_ids.append(sg_id)
         return sg_ids
 
+    @retries(max_attempts=3, wait=1)
```
e4r7hbug: Include an `exceptions=(boto3.exceptions.botocore.exceptions.ClientError)` in `retries` so we don't retry accidentally with `KeyError` or something else.

src/foremast/awslambda/awslambda.py
```diff
@@ -61,6 +61,22 @@ def _check_lambda(self):
...
+        Returns:
+            True if alias exists
+            False if alias does not exist
+        """
+        aliases = self.lambda_client.list_aliases(FunctionName=self.app_name)
+
+        for alias in aliases['Aliases']:
+            if alias['Name'] == self.env:
+                LOG.info('Found alias %s for function %s', self.env, self.app_name)
+                return True
```
e4r7hbug: Try to use one exit point.
```python
found = False

for alias in aliases['Aliases']:
    if alias['Name'] == self.env:
        found = True
        break

LOG.info('%s alias for function "%s" found? %s', self.env, self.app_name, found)
return found
```

src/foremast/awslambda/awslambda.py
```diff
@@ -61,6 +61,22 @@ def _check_lambda(self):
             pass
         return exists
 
+    def _check_lambda_alias(self):
+        """ Checks if lambda alias exists
```
e4r7hbug: Docstring should be like:
```python
"""Check if lambda alias exists.

Returns:
...
"""
```
`pydocstyle` can help with keeping docstrings consistent.

src/foremast/awslambda/awslambda.py
e4r7hbug: This function looks oddly similar to `foremast.utils.awslambda.get_lambda_alias_arn`. Can we refactor down to one function?

src/foremast/awslambda/awslambda.py
asibin: I am just curious why is this "more preferred" than having it the old way? For me at least old way is more readable, and easier to track. Also function is not really complex to warrant having only one exit point. I understand the point about multiple log statements but still think old way is more readable even though there are 2 log statements. I can change it, no problem, just trying to learn why it would be better :)

src/foremast/awslambda/awslambda.py
asibin: Thought about that while developing. LambdaFunction class presumes to have prop_path defined for a lot of its functions and handling missing prop_path would introduce more complex logic in handling that. 

We can call external function from `utils/` of course but then it would introduce "dependency" to another function (not that we don't have many already) and class itself won't reuse one API client. Since LambdaFunction will be executed more often (since it does the deploys) i think that having the atomic and "one-api-call" version is a good thing because each API call that we save at large scale will be a good thing.

I am against duplication of code but for me in this case it makes sense to have two similar functions. Of course, if you do like me to change it let me know which approach you would like.

asibin:
Fixed almost all styling / logic / test requirements. Left comments on those tasks i thought deserve comments :)

src/foremast/awslambda/awslambda.py
e4r7hbug: I was just thinking of this for consistency. We were starting to trend more toward one exit point. I'm okay with leaving this though, just thought I would mention.

src/foremast/awslambda/awslambda.py
e4r7hbug: We can refactor later.

tests/utils/test_awslambda.py
```diff
@@ -2,11 +2,40 @@
...
+import pytest
+
+from foremast.utils.awslambda import add_lambda_permissions, get_lambda_alias_arn
+from foremast.exceptions import LambdaAliasDoesNotExist
 
 ERROR_RESPONSE = {'Error': {}}
 
 
+def lambda_alias_list_mock():
+    return_dict = \
```
e4r7hbug: No backslashes.

PR #64, opened 2016-09-23
FIX: lambda trigger permissions now targeting function alias instead …

asibin:
…of app_name

### DIFF
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -22,7 +22,8 @@ import botocore
 from tryagain import retries
 
 from foremast.exceptions import InvalidEventConfiguration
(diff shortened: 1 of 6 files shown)


ndcampbell:
This looks good to me

e4r7hbug:
I think this is good too. Clear to merge.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -22,7 +22,8 @@
 from tryagain import retries
 
 from foremast.exceptions import InvalidEventConfiguration
-from foremast.utils import (get_details, get_env_credential, get_dns_zone_ids, update_dns_zone_record, get_properties, add_lambda_permissions)
+from foremast.utils import (get_details, get_env_credential, get_dns_zone_ids, update_dns_zone_record, get_properties,
```
asibin: Pycharm sorts this for me :) Actually, speaking of IDE it just warned me that we have unused imports in that file. New commit to fix that and i'll merge to avoid wasting time.

PR #69, opened 2016-09-28
Added lambda IAM permission to log by default

### LINKED ISSUE #68: Default lambda IAM role should include Cloudwatch
Attached lambda iam role is missing cloudwatch permissions which in turn doesn't allow for lambda logging which will create confusion with developers.

asibin:
Fixes #68


e4r7hbug:
Ah, good catch. Looks good for merge.

PR #71, opened 2016-09-29
Split tox into code / lint checks

sijis:
Its probably best to split code and linting tests. A linting check shouldn't necessarily fail a unit test.


e4r7hbug:
:+1:

PR #73, opened 2016-09-30
Feat/pipeline types

e4r7hbug:
Provide a way to restrict what deployment types are allowed. This allows for administrators to block certain types, like the upcoming `manual` type, if support is not desired.


ndcampbell:
LGTM 👍

Issue #72, opened 2016-09-30
feat: Manual deployment type


e4r7hbug:
Provide the ability to have foremast manage a predefined Pipeline using the raw JSON. This will be allow one to copy the JSON from the Pipeline, save it in a repository, and have foremast reconfigure the defined Pipeline with every run.

PR #74, opened 2016-09-30
feat: Provide centralized Git file retrieval

e4r7hbug:
Manual Pipelines will need to retrieve Pipeline JSON definition files from Git. This provides a concise way to grab the file and will be useful for refactoring `process_git_configs()` in [src/foremast/configs/prepare_configs.py](https://github.com/gogoair/foremast/blob/master/src/foremast/configs/prepare_configs.py).

See also: #72

### DIFF
--- a/src/foremast/utils/lookups.py
+++ b/src/foremast/utils/lookups.py
@@ -66,7 +66,7 @@ def ami_lookup(region='us-east-1', name='tomcat8'):
     return ami_id
 
 
-class GitLookup():
+class FileLookup():
     """Retrieve files from a local filesystem or remote GitLab Server.
 
     When _runway_dir_ is specified, the local directory is given priority and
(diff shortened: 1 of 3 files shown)


ndcampbell:
Should https://github.com/gogoair/foremast/blob/master/src/foremast/configs/prepare_configs.py#L32 be updated in this commit then to use the new util or do it in a different PR?

Also, will manual pipelines still allow passing in a "runway_dir" for the configuration? I don't want a hard gitlab dependency.

e4r7hbug:
I want to get the code into place first, then update the other code.

You're right about the `runway_dir`. I'll add support for that.

ndcampbell:
I think it should be generalized if it is not only using Git.

e4r7hbug:
Okay. I think this thing is done. Anyone want to give it one more look over?

sijis:
I'm merging this as its a precursor to other pending PRs.

PR #77, opened 2016-10-02
feat: ElasticSearch IAM template

### LINKED ISSUE #76: templates: ElasticSearch IAM statement
We need an ElasticSearch template in [src/foremast/templates/infrastructure/iam/](https://github.com/gogoair/foremast/tree/master/src/foremast/templates/infrastructure/iam). Use [src/foremast/templates/infrastructure/iam/dynamodb.json.j2](https://github.com/gogoair/foremast/blob/master/src/foremast/templates/infrastructure/iam/dynamodb.json.j2) as a base example. Try to restrict access to creating and deleting Resources. Remove the `for` loop if resource targeting is not needed.

Example AWS provided *AmazonESFullAccess* Policy.
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
      "es:*"
    ],
    "Effect": "Allow",
    "Resource": "*"
  }]
}
```

aaronrea:
Fixes #76 

I think this is what you are looking for. Doesn't make much sense (to me) to limit Resource: to a specific domain.

### DIFF
--- a/src/foremast/templates/infrastructure/iam/elasticsearch.json.j2
+++ b/src/foremast/templates/infrastructure/iam/elasticsearch.json.j2
@@ -1,0 +1,9 @@
+        {
+            "Sid": "ElasticSearchCreateDelete",
+            "Action": [
+                "es:CreateElasticsearchDomain",
+                "es:DeleteElasticsearchDomain"
+            ],
+            "Effect": "Allow",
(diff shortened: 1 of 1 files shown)


e4r7hbug:
Oooo, thank you for contributing! This indeed is what we need to start.

PR #78, opened 2016-10-03
feat: adding elasticache template

### LINKED ISSUE #75: templates: ElastiCache IAM statement
We need an ElastiCache template in [src/foremast/templates/infrastructure/iam/](https://github.com/gogoair/foremast/tree/master/src/foremast/templates/infrastructure/iam). Use [src/foremast/templates/infrastructure/iam/dynamodb.json.j2](https://github.com/gogoair/foremast/blob/master/src/foremast/templates/infrastructure/iam/dynamodb.json.j2) as a base example. Try to restrict access to creating and deleting Resources. Remove the `for` loop if resource targeting is not needed.

Example AWS provided *AmazonElastiCacheFullAccess* Policy.
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "elasticache:*",
      "Effect": "Allow",
      "Resource": "*"
    }
  ]
}
```

aaronrea:
Fixes #75 

I'm very well versed in elasticache but this follows what the docs list. The s3 bucket is used when restoring from a snapshot so I'm not sure if that would come in to play for how foremast uses these templates or if it is still needed if not restoring.

### DIFF
--- a/src/foremast/templates/infrastructure/iam/elasticache.json.j2
+++ b/src/foremast/templates/infrastructure/iam/elasticache.json.j2
@@ -1,0 +1,15 @@
+        {
+            "Sid": "ElasticacheFullAccess",
+            "Effect": "Allow",
+            "Action": [
+                "elasticache:*"
+            ],
+            "Resource": [
+            {%- for cluster_name in items %}
+                "arn:aws:elasticace:{{ region }}:{{ account_number }}:cluster/{{ cluster_name }}"
+                {%- if not loop.last -%}
+                ,
+                {%- endif -%}
+            {% endfor %} 
+            ]
+        }


aaronrea:
Just realized this one will need [ ] around it to allow multiple statements. Does foremast handle that? It might also be relevant to the s3 bucket template.

e4r7hbug:
Yup, we automatically wrap the statement in square brackets if the first `json.loads()` fails. See [src/foremast/iam/construct_policy.py](https://github.com/gogoair/foremast/blob/master/src/foremast/iam/construct_policy.py#L103-L106).

We do have the existing S3 template, [src/foremast/templates/infrastructure/iam/s3.json.j2](https://github.com/gogoair/foremast/blob/master/src/foremast/templates/infrastructure/iam/s3.json.j2), so we might want to remove the Statement you have here. People would need their `runway/pipeline.json` to have both Services defined:

```json
{
  "services": {
    "elasticache": true,
    "s3": ["bucket"]
  }
}
```

Would it be enough to add to the documentation ([_docs/configuration_files/pipeline_json/services.rest](https://github.com/gogoair/foremast/blob/master/_docs/configuration_files/pipeline_json/services.rest)) that S3 should be added when using ElastiCache? Or does it make sense to have the ElastiCache template combine both?

We'll double check with @fadi-almasri to see if the ElastiCache portion aligns with what he wants.

fadi-almasri:
I would suggest giving all permissions (i.e: elasticache:*) for a specific resource that the developer is interested in managing. So, we expect them to provide the resource ARN in pipeline.json.

aaronrea:
Sorry, meant to say "not very well versed in elasticache". I'll get this changed up over the weekend. Thanks for the feedback.

e4r7hbug:
Haha, how much a small word can make, no worries! Thanks for helping us out.

aaronrea:
Ok I think this is more what you're looking for. Sorry if there is any confusion, I messed up a rebase on my local and had to fix. 

The S3 bucket is optional and only applies if you a restoring a redis cluster snapshot specifically. Maybe that warrants a different template/more logic. Do even need to account for this? It seems we are assuming a pre-existing elasticache resource.

e4r7hbug:
Thanks @aaronrea, we'll take this. I'll get this merged in later today.

PR #79, opened 2016-10-04
fix: Expand certificate paths

e4r7hbug:
For some reason we weren't resolving the paths.


jvasallo:
LGTM! Thanks for fixing this! 👍

PR #81, opened 2016-10-04
Updated renumerate function to remove hardcoded stage names

### LINKED ISSUE #58: Refactor renumerate_stages
https://github.com/gogoair/foremast/blob/master/src/foremast/pipeline/renumerate_stages.py

This can be refactored to be more agnostic to stage names. Right now it is kind of specific to the Gogo workflow but there should be a better way.

ndcampbell:
fixes #58 

I changed the logic to look for `"a"` or `"b"`  in the stage templates. Stages marked with `"a"` are mainline required pipeline stages. stages marked with `"b"` are secondary parallel stages. The order is already handled by https://github.com/gogoair/foremast/blob/master/src/foremast/templates/pipeline/pipeline_stages.json.j2 template.

This change would break existing templates, but makes the enumerate code a lot cleaner and supportable.

I tested this against our current pipeline and it works consistently.

### DIFF
--- a/src/foremast/pipeline/renumerate_stages.py
+++ b/src/foremast/pipeline/renumerate_stages.py
@@ -36,14 +36,14 @@ def renumerate_stages(pipeline):
 
     main_index = 0
     for stage in stages:
-        if stage['refId'] == 'a':
+        if stage['refId'].lower() == 'master':
             if main_index == 0:
                 stage['requisiteStageRefIds'] = []
             else:
                 stage['requisiteStageRefIds'] = [str(main_index)]
             main_index += 1
             stage['refId'] = str(main_index)
-        elif stage['refId'] == 'b':
+        elif stage['refId'].lower() == 'branch':
             stage['refId'] = str(main_index*100)
             stage['requisiteStageRefIds'] = [str(main_index)]
 
--- a/src/foremast/pipeline/renumerate_stages.py
+++ b/src/foremast/pipeline/renumerate_stages.py
@@ -22,9 +22,13 @@ LOG = logging.getLogger(__name__)
 
 def renumerate_stages(pipeline):
     """Renumber Pipeline Stage reference IDs to account for dependencies.
-       stage order is defined in the templates.
-       refId == 'a' - Any mainline required stage
-       refId == 'b' - Any secondary stage that should be ran in parallel
+
+    stage order is defined in the templates. The ``refId`` field dictates
+    if a stage should be mainline or parallel to other stages.
+
+        * ``master`` - A mainline required stage. Other stages depend on it
+        * ``branch`` - A stage that should be ran in parallel to master stages.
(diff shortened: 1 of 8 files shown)


sijis:
I think it all makes sense. It would be more clear if `"a"` and `"b"` was something like `"required"` and `"optional"`. We will be sad 6 months from now and what what does "a" or "b" mean.

ndcampbell:
I agree but am having a hard time with what key to use. `"optional"` is not the right word, the stages are still required, just no other stages depend on them. Maybe `"mainline"` and `"parallel"`?

e4r7hbug:
Could go with Git naming conventions: master, merge, branch.

ndcampbell:
Yeah, I kind of like `"master"`, `"branch"`, and `"merge"`. They are at least a good visual representation of the intended goal.

sijis:
@ndcampbell @e4r7hbug I do like them with exception of `merge`. I don't think that's as clear as `master` and `branch`. I don't have an alternative naming, so go with that.

src/foremast/pipeline/renumerate_stages.py
```diff
@@ -57,24 +34,18 @@ def renumerate_stages(pipeline):
...
-            stage['requisiteStageRefIds'] = [str(main_index)]
-            stage['refId'] = str(main_index * 102)
-        elif stage['type'] == 'bake' or idx == 0:
-            stage['requisiteStageRefIds'] = []
-            stage['refId'] = str(main_index)
-        else:
-            stage['requisiteStageRefIds'] = [str(main_index)]
+    main_index = 0
+    for stage in stages:
+        if stage['refId'].lower() == 'master':
```
e4r7hbug: Save `stage['refId'].lower()` in a variable.

src/foremast/pipeline/renumerate_stages.py
```diff
@@ -22,32 +22,9 @@
...
-    +---------+------------------------+------------+
-    | 4       | judgement              | 3          |
-    +---------+------------------------+------------+
-    | 5       | deploy stage           | 4          |
-    +---------+------------------------+------------+
-    | 6       | QE stage               | 5          |
-    +---------+------------------------+------------+
-    | 500     | git tagger stage       | 5          |
-    +---------+------------------------+------------+
+       stage order is defined in the templates.
```
e4r7hbug: Docstring formatting.
```python
"""Renumber....

Stage order is defined in the templates. Use ``refId`` to determine dependency ordering.

    * ``a`` - Any...
    * ``b`` - Any...
"""
```

src/foremast/pipeline/renumerate_stages.py
```diff
@@ -22,32 +22,9 @@
...
+       refId == 'a' - Any mainline required stage
```
e4r7hbug: Update `a` and `b` to the new names we're trying out.

PR #82, opened 2016-10-04
refactor: Use new central FileLookup

e4r7hbug:
Switch local and remote file lookups to use the centralized Class `FileLookup` in #74.

See also: #72


ndcampbell:
LGTM. As far as I can tell, it does not change any functionality or configs, correct? Just centralizes around one until for getting the configs.

e4r7hbug:
Yup, I left the current logic in place, so any Bugs in there are still there.

PR #83, opened 2016-10-04
feat: Manual Pipeline support

e4r7hbug:
When `{"type": "manual"}`, use the listed files in `{"pipeline_files": []}` to create Pipelines.

See also: #74
Closes: #72

### DIFF
--- a/_docs/configuration_files/pipeline_json/index.rst
+++ b/_docs/configuration_files/pipeline_json/index.rst
@@ -19,6 +19,7 @@ Example Configuration
 Configuration Details
 ----------------------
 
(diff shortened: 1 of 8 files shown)


ndcampbell:
LGTM.

Worth noting I do not think this would work internally for us since we do not pass the runway_dir, we get the files directly from Gitlab. We would need to revamp the Jenkins job to allow an optional runway_dir path for manual pipelines.

ndcampbell:
@e4r7hbug Anything else needed on this?

e4r7hbug:
Nothing else that I can think of. I'll merge this in.

Issue #70, opened 2016-09-29
templates: Kinesis Firehose IAM statement


e4r7hbug:
We need a Kinesis Firehose template in [src/foremast/templates/infrastructure/iam/](https://github.com/gogoair/foremast/tree/master/src/foremast/templates/infrastructure/iam). Use [src/foremast/templates/infrastructure/iam/dynamodb.json.j2](https://github.com/gogoair/foremast/blob/master/src/foremast/templates/infrastructure/iam/dynamodb.json.j2) as a base example to support a list of Delivery Streams.

Example AWS provided *AmazonKinesisFirehoseFullAccess* Policy.
```json
{
  "Version": "2012-10-17",
  "Statement": [
  {
    "Action": [
    "firehose:*"
  ],
  "Effect": "Allow",
  "Resource": "*"
  }]
}
```

PR #84, opened 2016-10-04
feat: Kinesis Firehose IAM template

e4r7hbug:
There seems to be a lot of things associated with Firehose, so any help with understanding what else would be required will be appreciated.

Fixes: #70


e4r7hbug:
Merging in for now. We can make changes as we start using it.

PR #86, opened 2016-10-11
feat: Unified CLI commands

e4r7hbug:
This provides a more explorable CLI structure. Try to ignore the configuration spew for now, we'll need to clean up the side effect of importing `foremast.consts` eventually.

``` bash
$ foremast
No configuration found in the following locations:

/etc/foremast/foremast.cfg
/home/ntangsurat/.foremast/foremast.cfg
./.foremast/foremast.cfg

Section missing from configurations: [base]
[base] missing key "gate_api_url", using None.
[base] missing key "git_url", using None.
[base] missing key "domain", using 'example.com'.
[base] missing key "envs", using ''.
[base] missing key "regions", using ''.
[base] missing key "types", using 'ec2,lambda'.
[base] missing key "templates_path", using None.
[base] missing key "ami_json_url", using None.
[base] missing key "default_ec2_securitygroups", using ''.
[base] missing key "default_elb_securitygroups", using ''.
Section missing from configurations: [credentials]
[credentials] missing key "gitlab_token", using None.
[credentials] missing key "slack_token", using None.
Section missing from configurations: [whitelists]
[whitelists] missing key "asg_whitelist", using ''.
[base] missing key "gate_client_cert", using ''.
[base] missing key "gate_ca_bundle", using ''.
usage: foremast [-h] [-d] [-s] {infra,pipeline,rebuild,autoscaling} ...

Foremast, your ship's support.

optional arguments:
  -h, --help            show this help message and exit
  -d, --debug           Set DEBUG output (default: 20)
  -s, --short-log       Truncated logging format (default: %(asctime)s
                        [%(levelname)s] %(name)s:%(funcName)s:%(lineno)d -
                        %(message)s)

Commands:
  Available activies

  {infra,pipeline,rebuild,autoscaling}
    infra               Entry point for preparing the infrastructure in a
                        specific env.
    pipeline            Pipeline subcommands.
    rebuild             Entry point for rebuilding pipelines. Use to rebuild
                        all pipelines or a specific group.
    autoscaling         Create Auto Scaling Policy for an Auto Scaling Group.
```

PR #87, opened 2016-10-11
feat: Tester subcommands

e4r7hbug:
Provide some help with checking a Spinnaker setup from the Foremast perspective. This should provide a nicer way of seeing which Foremast configurations are not set properly. Also, this could help with detecting if components of Spinnaker are down.

Depends on: #86

``` bash
$ foremast tester -h
No configuration found in the following locations:

/etc/foremast/foremast.cfg
/home/ntangsurat/.foremast/foremast.cfg
./.foremast/foremast.cfg

Section missing from configurations: [base]
[base] missing key "gate_api_url", using None.
[base] missing key "git_url", using None.
[base] missing key "domain", using 'example.com'.
[base] missing key "envs", using ''.
[base] missing key "regions", using ''.
[base] missing key "types", using 'ec2,lambda'.
[base] missing key "templates_path", using None.
[base] missing key "ami_json_url", using None.
[base] missing key "default_ec2_securitygroups", using ''.
[base] missing key "default_elb_securitygroups", using ''.
Section missing from configurations: [credentials]
[credentials] missing key "gitlab_token", using None.
[credentials] missing key "slack_token", using None.
Section missing from configurations: [whitelists]
[whitelists] missing key "asg_whitelist", using ''.
[base] missing key "gate_client_cert", using ''.
[base] missing key "gate_ca_bundle", using ''.
usage: foremast tester [-h] {all,gate} ...

optional arguments:
  -h, --help  show this help message and exit

Testers:
  {all,gate}
    all       Run all tests. Args: args (CliArgs): Arguments from CLI.
    gate      Check Gate connection.
```

### DIFF
--- a/src/foremast/__main__.py
+++ b/src/foremast/__main__.py
@@ -61,7 +61,13 @@ def add_tester(subparsers):
     """Test Spinnaker setup."""
     tester_parser = subparsers.add_parser(
         'tester', help=add_tester.__doc__, formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-    tester_parser.set_defaults(func=tester.all_tests)
+    tester_parser.set_defaults(func=tester_parser.print_help)
+
+    tester_subparsers = tester_parser.add_subparsers(title='Testers')
+
+    tester_all_parser = tester_subparsers.add_parser(
+        'all', help=tester.all_tests.__doc__, formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    tester_all_parser.set_defaults(func=tester.all_tests)
 
 
 def main(manual_args=None):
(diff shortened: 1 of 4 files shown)


sijis:
Is the idea to add other components/checks to this in the future besides gate?

I'm on the fence about this and maybe its because its called a 'Tester'.
I think in reality this is a 'Credential' or 'Connectivity' validator.

e4r7hbug:
Yeah, the idea would be to check if `foremast` can reach out to the different Spinnaker components. Should help clarify if your configuration is correct. Hm…I wouldn't be opposed to change the command name.

sijis:
@e4r7hbug I think 'Validate' could be the better name.
Eventually we could add a config validator among other minor things. I'm visualizing the use as
`$ foremast validate config`

@ndcampbell thoughts?

e4r7hbug:
`validate` makes sense. So…now we should probably change things to consistently say `validate` instead of `validator`. Because every time I type "validator", there's a red squiggly.

sijis:
I made the suggested changes.

e4r7hbug:
Good stuffs.

PR #89, opened 2016-10-12
Fixed line lengths to be 120 chars for import statements using isort

jvasallo:
Noticed that we should be using isort and we use 120 length line length on our project. This is just a small lint fixing this in the imports.


ndcampbell:
Cool, looks good to me.

PR #91, opened 2016-10-18
Updated lambda policy after AWS addressed a bug in their documentation:

jvasallo:
Amazon has mentioned there was an issue in their documentation:

See the new docs here.
http://docs.aws.amazon.com/lambda/latest/dg/policy-templates.html#LambdaVPCAccessExecutionRole


sijis:
:+1:

PR #93, opened 2016-10-21
Changed pipeline_id to be None if no existing pipeline exists.

ndcampbell:
This fixes an issue where the empty string causes problems in S3.

It tested this against an S3 and Cassanada backend for Front50. Keeping the value None instead of an empty string works in both.


sijis:
I see no issue with this.

Its good to know that this will fix for cassandra and s3 backends.

:+1:

PR #94, opened 2016-10-24
changed default_provider to discovery

ndcampbell:
This was `Amazon` for Eureka deployments but this causes it not to check Eureka healthchecks, only Amazon. Changed default to `Discovery` and this solves the issues. ELB deployments work the same as before.


sijis:
:+1:

PR #96, opened 2016-10-28
fix: Handle missing env

e4r7hbug:
When `pipeline.json` is configured with the wrong environments or the Security Group definition is missing in the `application.json`, we should alert with a more informative error.


jvasallo:
Looks like a good fix to me, more alerting is always good and this has traditionally been a rather ambiguous error for our devs. Nice!

PR #98, opened 2016-11-01
Updated tests to use the tests/ directory specifically

jvasallo:
noticed when tox was running, it was attempting to run all the tests in my local virtualenv. Fixed it to just do foremast specific tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ pep8maxlinelength = 120
 
 [testenv]
 deps = -rrequirements-dev.txt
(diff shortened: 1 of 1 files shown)


sijis:
This was initially there when we had out linting and tests under the same tasks. In that scenario we needed to ensure linting was checked against other files in the root of the project.

We've since moves linting as a separate task, so this could probably be added again.

e4r7hbug:
So that Joel and others using a virtual environment in the cloned repository can continue to contribute, we'll get this merged in.

PR #99, opened 2016-11-07
Fixed typo in firehose policy template

jvasallo:
There was a typo in the policy for firehose (it was granting dynamo not firehose)


fadi-almasri:
Looks good.

PR #101, opened 2016-11-14
Ensure threshold and period_minutes are integers

### LINKED ISSUE #100: Multiplying a string
If someone has the following asg policy:
```
  "asg": {
    "subnet_purpose": "internal",
    "min_inst": 1,
    "max_inst": 5,
    "scaling_policy": {
      "metric": "NetworkIn",
      "threshold": "20000",
      "period_minutes": 1,
      "statistic": "Minimum"
    }
  },
```
You will see an error, like so:
```
packages/foremast/autoscaling_policy/create_policy.py", line 79, in prepare_policy_template
    'asg']['scaling_policy']['threshold'] * 0.5
TypeError: can't multiply sequence by non-int of type 'float'
```

The reason is that threshold is a string. If threshold is an int, this works.

We should ensure that thresholds are typecast to `int()`, as a possible fix for this issue.

sijis:
This handles cases where those keys are set as strings
even though we expect integers.

Fixes #100

### DIFF
--- a/src/foremast/autoscaling_policy/create_policy.py
+++ b/src/foremast/autoscaling_policy/create_policy.py
@@ -75,8 +75,8 @@ class AutoScalingPolicy:
             template_kwargs['scalingAdjustment'] = 1
 
         elif scaling_type == 'scale_down':
-            self.settings['asg']['scaling_policy']['threshold'] = self.settings[
-                'asg']['scaling_policy']['threshold'] * 0.5
+            self.settings['asg']['scaling_policy']['threshold'] = int(
+                self.settings['asg']['scaling_policy']['threshold']) * 0.5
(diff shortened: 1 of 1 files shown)


ndcampbell:
Why just checking for thresholds? If we enforce type for the other fields I think we should enforce it for thresholds.

sijis:
@ndcampbell We could absolutely do it. The primary type of the majority of the fields are strings, not int.

Temporarily we could move forward with this, however, we'd have to find a "better" way to ensure the types match. One thing I bumped into recently was: http://docs.python-cerberus.org/en/stable/. This look interesting and could be useful to validate type's sent.

jvasallo:
This looks good to me. Thanks for the PR.

PR #103, opened 2016-11-21
[WIP] Lambda should respect defined instance profile

sijis:
This should allow a custom defined instance_profile to work in lambda pipelines.

/cc @sbasgall

### DIFF
--- a/src/foremast/awslambda/awslambda.py
+++ b/src/foremast/awslambda/awslambda.py
@@ -60,7 +60,9 @@ class LambdaFunction(object):
         self.role = app.get('lambda_role') or generated.iam()['lambda_role']
         self.timeout = app['lambda_timeout']
 
-        self.role_arn = get_role_arn(self.role, self.env, self.region)
+        self.role_arn = get_role_arn(
+            self.properties[env]['app'].get('instance_profile', generated.iam()['role']),
+            self.env, self.region)
 
         self.session = boto3.Session(profile_name=self.env, region_name=self.region)
         self.lambda_client = self.session.client('lambda')


src/foremast/awslambda/awslambda.py
```diff
@@ -57,7 +57,9 @@ def __init__(self, app, env, region, prop_path):
         self.memory = self.properties[env]['app']['lambda_memory']
         self.timeout = self.properties[env]['app']['lambda_timeout']
 
-        self.role_arn = get_role_arn(generated.iam()['role'], self.env, self.region)
+        self.role_arn = get_role_arn(
+            self.properties[env]['app'].get('instance_profile', generated.iam()['role']),
```
e4r7hbug: `profile` and `role` are two different words. I predict there will be an issue here. We must double check.

sijis: That's a good point. I'll further look into it.

src/foremast/awslambda/awslambda.py
```diff
@@ -60,7 +60,9 @@ def __init__(self, app, env, region, prop_path):
         self.role = app.get('lambda_role') or generated.iam()['lambda_role']
         self.timeout = app['lambda_timeout']
 
-        self.role_arn = get_role_arn(self.role, self.env, self.region)
+        self.role_arn = get_role_arn(
+            self.properties[env]['app'].get('instance_profile', generated.iam()['role']),
```
e4r7hbug: It doesn't look like `self.role` gets used anywhere. Save this long line to `self.role` and use that instead.

PR #104, opened 2016-12-01
fix: Include api details in docs

sijis:
Fix docs so they include api details within the 'Foremast Modules" section.


ndcampbell:
LGTM

PR #105, opened 2016-12-02
Feature/multi region dns

ndcampbell:
Added the functionality to create Failover DNS when multiple regions are specified. I think this could use a really heavy review but the functionality is all there.


jvasallo:
Taking a look at this now! Thanks for the work on this.

jvasallo:
Looks good, lets file an issue for documentation around this! Thanks!

PR #106, opened 2016-12-05
added purpose to get_subnet for elbs

ndcampbell:
I think this was a bug, get_subnets() only ever looked at internet subnets regardless of what was configured.


e4r7hbug:
Looks good.

PR #108, opened 2016-12-13
default region specific DNS

ndcampbell:
made it so that when DNS is created the global record and region specific DNS is created. This is configurable but true by default

### DIFF
--- a/src/foremast/runner.py
+++ b/src/foremast/runner.py
@@ -158,6 +158,7 @@ class ForemastRunner(object):
         elb_subnet = self.configs[self.env]['elb']['subnet_purpose']
         regions = self.configs[self.env]['regions']
         failover = self.configs[self.env]['dns']['failover_dns']
+        regionspecific_dns = self.configs[self.env]['dns']['region_specific']
 
         dnsobj = dns.SpinnakerDns(app=self.app,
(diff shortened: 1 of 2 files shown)


src/foremast/runner.py
```diff
@@ -170,6 +171,8 @@ def create_dns(self):
             dnsobj.create_failover_dns(primary_region=primary_region)
         else:
             dnsobj.create_elb_dns(regionspecific=False)
+            if regionspecific_dns: #If true, Also create a region specific DNS record
+                dnsobj.create_elb_dns(regionspecific=True)
```
jvasallo: Should we not create the region specific first, then create the parent record?
internal-elb1212.amazon.com
something.cool.us-east-1.staging.company.org
something.cool.staging.company.org

ndcampbell: Why does it matter? It is creating `app.group.region.env.domain.com` and `app.group.env.domain.com`. Both of which are CNAMEs to the same ELB record.

jvasallo: Ah makes sense for now. Both records point to the ELB by default. Sounds fine by me for now.

jvasallo:
Looks good to me!

PR #109, opened 2016-12-15
fix: Manual pipeline safe name

e4r7hbug:
Need to remove bad characters from the generated Pipeline name in accordance with Spinnaker's name checker. This allows manual Pipelines to be created and deleted properly.

### DIFF
--- a/src/foremast/pipeline/create_pipeline_manual.py
+++ b/src/foremast/pipeline/create_pipeline_manual.py
@@ -14,6 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 """Create manual Pipeline for Spinnaker."""
+from ..utils import normalize_pipeline_name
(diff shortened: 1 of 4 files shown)


sijis:
This needs a simple test for the normalized function.

e4r7hbug:
Test case added.

sijis:
👍 I'm good with this.

e4r7hbug:
Turns out, updating an existing Pipeline without deleting it is fairly trivial with all our existing functions.

sijis:
I would have moved the pipeline removal to a separate PR. It may be trivial but I would like more eyes on that specific change.

e4r7hbug:
Split Pipeline deletion related Commits to #110.

sijis:
Thanks for breaking that up into 2 parts.

PR #111, opened 2016-12-20
Add application name in successful pipeline message

sijis:
Include the application name in successful pipeline creation message.


e4r7hbug:
Key access should be safe since everyone using the wrapper template has `application` and `name` set. I think this can get merged in.

ndcampbell:
LGTM

PR #114, opened 2017-01-06
Added logic for finding and deleting old CNAME records

ndcampbell:
This is necessary for when applications change to multiregion. An upsert does not work since the record type changes.

### DIFF
--- a/src/foremast/utils/dns.py
+++ b/src/foremast/utils/dns.py
@@ -108,12 +108,12 @@ def delete_existing_cname(env, zone_id, dns_name):
         HostedZoneId=zone_id
     )
     for item in response['ResourceRecordSets']:
-        if item['Name'][:-1] == newrecord_name and item['Type'] == 'CNAME':
+        if item['Name'].rstrip('.') == newrecord_name and item['Type'] == 'CNAME':
             startrecord = item
             LOG.info("Found old record: %s", item)
             break
 
-    if oldrecord:
+    if startrecord:
         LOG.info("Deleting old record: %s", newrecord_name)
         del_response = client.change_resource_record_sets(
             HostedZoneId=zone_id,
--- a/src/foremast/utils/dns.py
+++ b/src/foremast/utils/dns.py
@@ -91,7 +91,7 @@ def update_dns_zone_record(env, zone_id, **kwargs):
 
     LOG.debug('Route53 JSON Response: \n%s', pformat(response))
 
-def delete_existing_cname(env, zone_id, **kwargs):
+def delete_existing_cname(env, zone_id, dns_name):
     """Function to delete an existing CNAME record. This is
     used when updating to multi-region for deleting old records.
     The record can not just be upserted since it changes types.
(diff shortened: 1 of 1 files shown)


src/foremast/utils/dns.py
```diff
@@ -91,6 +91,43 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+def delete_existing_cname(env, zone_id, **kwargs):
+    """Function to delete an existing CNAME record. This is
+    used when updating to multi-region for deleting old records.
+    The record can not just be upserted since it changes types.
+
+    Args:
+        env (str): Deployment environment.
+        zone_id (str): Route53 zone id.
+
+    Keyword Args:
```
e4r7hbug: All arguments should go under `Args:`.

src/foremast/utils/dns.py
```diff
@@ -91,6 +91,43 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+
+    """
+    oldrecord = None
+    recordname = kwargs.get('dns_name')
+    client = boto3.Session(profile_name=env).client('route53')
+    response = client.list_resource_record_sets(
+        HostedZoneId=zone_id
+    )
+    for item in response['ResourceRecordSets']:
+        if item['Name'][:-1] == recordname and item['Type'] == 'CNAME':
```
e4r7hbug: Can we do `.startswith()` instead?

src/foremast/utils/dns.py
```diff
@@ -91,6 +91,43 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+            oldrecord = item
+            LOG.info("Found Old Record: %s", item)
+            break
+
+    if oldrecord:
+        LOG.info("Deleteing Old Record: %s", recordname)
```
e4r7hbug: Spell check.

src/foremast/utils/dns.py
```diff
@@ -91,6 +91,43 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+        zone_id (str): Route53 zone id.
+
+    Keyword Args:
+        dns_name (str): FQDN of application's dns entry to add/update.
+
+    """
+    oldrecord = None
+    recordname = kwargs.get('dns_name')
+    client = boto3.Session(profile_name=env).client('route53')
+    response = client.list_resource_record_sets(
```
e4r7hbug: Give a `StartRecordName` and `StartRecordType` to get to the point.

src/foremast/utils/dns.py
e4r7hbug: Actually, `[:-1]` doesn't look like it's needed.

sijis:
Is the assumption made that all apps are muti-region?

ndcampbell:
@sijis, No. It is only assumed this is needed if two regions are listed in the application.json file.

src/foremast/utils/dns.py
ndcampbell: [:-1] is be cause there is a period at the end of `item["Name"]`. I could just use start with I guess but this felt more accurate to me

src/foremast/utils/dns.py
```diff
@@ -91,6 +91,40 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+
+    Args:
+        env (str): Deployment environment.
+        zone_id (str): Route53 zone id.
+        dns_name (str): FQDN of application's dns entry to add/update.
+    """
+    startrecord = None
+    newrecord_name = dns_name
+    client = boto3.Session(profile_name=env).client('route53')
+    response = client.list_resource_record_sets(
```
e4r7hbug: O wait, I meant give `list_resource_record_sets` the `StartRecordName` and `StartRecordType` keyword arguments.

src/foremast/utils/dns.py
```diff
@@ -91,6 +91,40 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+        HostedZoneId=zone_id
+    )
+    for item in response['ResourceRecordSets']:
+        if item['Name'][:-1] == newrecord_name and item['Type'] == 'CNAME':
```
e4r7hbug: Can you do `old_record_name = item['name'].rstrip('.')`? That would be clearer to me.

src/foremast/utils/dns.py
```diff
@@ -91,6 +91,40 @@ def update_dns_zone_record(env, zone_id, **kwargs):
...
+            startrecord = item
+            LOG.info("Found old record: %s", item)
+            break
+
+    if oldrecord:
```
e4r7hbug: Doesn't look like `oldrecord` is declared anywhere.

src/foremast/utils/dns.py
ndcampbell: You are right, that is clearer and less bug prone. Good idea.

src/foremast/utils/dns.py
ndcampbell: Oh, I dont really need to break them out IMO because I only ever look at them once for the if statement. I just take the whole record and put it in my `change_resource_record_sets` but I never specifically to look at Name and Type again.

src/foremast/utils/dns.py
ndcampbell: Oh thanks!

ndcampbell:
I am going to go ahead and merge this in, functionally it works. If you have any issues with the conventions I used feel free to put in another PR. There is also some more work that will be needed for this, like what happens if you switch back to single region? Fixes for another PR

PR #115, opened 2017-01-06
added emr policy to templates

ndcampbell:
added ElasticMapReduce permissions to our policy templates


e4r7hbug:
Looks good. Too bad AWS doesn't have a way of restricting the resources at the moment.

PR #118, opened 2017-01-17
Feature/cookie stickiness

ndcampbell:
Added the ability to set app cookie stickiness on an ELB listener.

Will create an issue to add load balancer cookie and time based stickiness as well but this PR just covers application created cookies.


jvasallo:
Looks good to me, we even tested the edge scenario of having leveraged the ELB policies section for cipher policies. This maintained sufficient backwards compatibility and shouldn't break users using that feature. Thanks for testing that as well.!

PR #121, opened 2017-01-18
Make sure primary record exists before creating secondary

ndcampbell:
Added a lookup to make sure that the primary record exists before creating the secondary failover record. I also updated the CNAME lookup code to use paginators to solve a future issue.

This solves the issue we had with the failover record being created before the primary record and bringing down the service temporarily.


src/foremast/utils/dns.py
```diff
@@ -148,6 +149,13 @@ def update_failover_dns_record(env, zone_id, **kwargs):
     zone_name = hosted_zone_info['HostedZone']['Name'].rstrip('.')
     dns_name = kwargs.get('dns_name')
 
+    #Check that the primary record exists
+    failover_state = kwargs.get('failover_state')
+    if failover_state.lower() != 'primary':
```
sijis: Could this be combined to say
```
if failover_state.lower() != 'primary' and not primary_record_exists(env, zone_id, dns_name):
    raise PrimaryDNSRecordNotFound("Primary Failover DNS record not found: {}".format(dns_name))
```

sijis: Actually, nevermind, the `primary_record_exists()` is an necessary, if `failover_state` isn't primary. Its also makes an extra api call which could be expensive and unwarranted in this scenario.

src/foremast/utils/dns.py
```diff
@@ -104,14 +105,14 @@ def delete_existing_cname(env, zone_id, dns_name):
...
-    )
-    for item in response['ResourceRecordSets']:
-        if item['Name'].rstrip('.') == newrecord_name and item['Type'] == 'CNAME':
-            startrecord = item
-            LOG.info("Found old record: %s", item)
-            break
+    pager = client.get_paginator('list_resource_record_sets')
+
+    for rset in pager.paginate(HostedZoneId=zone_id):
+        for item in rset['ResourceRecordSets']:
```
sijis: I would probably change variable from `item` to `record` as in the `primary_record_exists()` method. I think it makes it clearer.

src/foremast/utils/dns.py
```diff
@@ -148,6 +149,13 @@ def update_failover_dns_record(env, zone_id, **kwargs):
     zone_name = hosted_zone_info['HostedZone']['Name'].rstrip('.')
     dns_name = kwargs.get('dns_name')
 
+    #Check that the primary record exists
+    failover_state = kwargs.get('failover_state')
+    if failover_state.lower() != 'primary':
+        primary_exists = primary_record_exists(env, zone_id, dns_name)
+        if not primary_exists:
+            raise PrimaryDNSRecordNotFound("Primary Failover DNS record not found: {}".format(dns_name))
```
sijis: Do you think its worth creating the primary record at this point since one does not exist?

ndcampbell: No, the issue with that is you are not guaranteed that the primary ELB even exists when creating the secondary record. By just failing instead of creating the primary, I think it avoids some weird corner case issues that would get hacky to work around. I'd also rather keep the runs 'pure', meaning they only impact the region you are actively deploying too. 

I am open to debate on this, those are just my justifications for doing it this way.

src/foremast/utils/dns.py
```diff
@@ -104,14 +105,14 @@ def delete_existing_cname(env, zone_id, dns_name):
...
-            LOG.info("Found old record: %s", item)
-            break
+    pager = client.get_paginator('list_resource_record_sets')
+
+    for rset in pager.paginate(HostedZoneId=zone_id):
+        for record in rset['ResourceRecordSets']:
+            if record['Name'].rstrip('.') == newrecord_name and record['Type'] == 'CNAME':
+                startrecord = record
+                LOG.info("Found old record: %s", record)
+                break
```
e4r7hbug: This block has minor differences from `primary_record_exists()`. Maybe make that function slightly more generic and use that here.

ndcampbell:
@e4r7hbug I took your advice and abstracted out that "find records" bit. Can you give it another review? 

I am also writing a test case and will have that in before I merge.

src/foremast/utils/dns.py
```diff
@@ -91,6 +92,30 @@ def update_dns_zone_record(env, zone_id, **kwargs):
 
     LOG.debug('Route53 JSON Response: \n%s', pformat(response))
 
+def find_existing_record(env, zone_id, dns_name, valuecheck=[]):
```
e4r7hbug: Can you split `valuecheck` into `check_type` and `check_value` keyword arguments instead? I'm not fond of passing 2-tuples around.

PR #122, opened 2017-01-27
mocked config values so that it works with local foremast configs

ndcampbell:
I was having to rename my local config for tests to work with default values. I mocked these defaults and now tests still work with out changing local foremast configs.

### DIFF
--- a/tests/dns/test_dns.py
+++ b/tests/dns/test_dns.py
@@ -32,7 +32,6 @@ def test_dns_creation(mock_get_details, mock_properties, mock_find_elb, mock_dns
     }
 
     # mock results
-    #mock_domain.value = 'example.com'
     mock_get_details.return_value.app_name.return_value = 'myapp'
     mock_get_details.return_value.dns.return_value = dns_elb
     mock_properties.return_value = {'dns': {'ttl': 60}}
--- a/tests/dns/test_dns.py
+++ b/tests/dns/test_dns.py
@@ -22,6 +22,7 @@ from unittest.mock import patch, MagicMock
 @patch('foremast.dns.create_dns.get_dns_zone_ids')
 @patch('foremast.dns.create_dns.find_elb')
 @patch('foremast.dns.create_dns.get_properties')
+@patch('foremast.dns.create_dns.DOMAIN', 'example.com')
 @patch('foremast.dns.create_dns.get_details')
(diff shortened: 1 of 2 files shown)


tests/dns/test_dns.py
```diff
@@ -31,6 +32,7 @@ def test_dns_creation(mock_get_details, mock_properties, mock_find_elb, mock_dns
     }
 
     # mock results
+    #mock_domain.value = 'example.com'
```
e4r7hbug: Remove comment.

tests/iam/test_iam_construct.py
```diff
@@ -68,9 +72,10 @@ def test_construct_cloudwatchlogs(requests_get):
 
 @mock.patch('foremast.utils.credentials.API_URL', 'http://test.com')
 @mock.patch('foremast.utils.credentials.requests.get')
-def test_construct_s3(requests_get):
+@mock.patch('foremast.utils.templates.TEMPLATES_PATH', None)
```
e4r7hbug: Remove `patch` since `fixture` is working.

tests/iam/test_iam_construct.py
```diff
@@ -95,9 +100,12 @@ def test_construct_s3(requests_get):
 
 @mock.patch('foremast.utils.credentials.API_URL', 'http://test.com')
 @mock.patch('foremast.utils.credentials.requests.get')
-def test_construct_s3_buckets(requests_get):
+@mock.patch('foremast.utils.templates.TEMPLATES_PATH', None)
```
e4r7hbug: Remove `patch` since `fixture` is working.

tests/iam/test_iam_construct.py
```diff
@@ -95,9 +100,12 @@ def test_construct_s3(requests_get):
...
+def test_construct_s3_buckets(requests_get, get_base_settings):
     """Check S3 Policy with multiple Buckets listed."""
-    pipeline_settings = copy.deepcopy(BASE_SETTINGS)
+    #print(get_base_settings)
+    #pipeline_settings = copy.deepcopy(BASE_SETTINGS)
```
e4r7hbug: Remove comments.

tests/iam/test_iam_construct.py
ndcampbell: These are necessary. The fixure and the test needs the patch

Issue #95, opened 2016-10-27
Disable / Rework environment specific rules


sijis:
In the following code below we are ensuring that the 'dev' environment has a few specific settings. Now, this applies to gogo but may not be a good idea for other users to have the same restrictions. Other foremast users probably have an account named 'dev' and are likely wondering why those asg min/max settings aren't adhered to.

The code in question is:
https://github.com/gogoair/foremast/blob/master/src/foremast/pipeline/construct_pipeline_block.py#L135-L143

We need to figure out a better way to allow this functionality or remove this stuff completely.

ndcampbell:
Good catch.

It looks like this logic could be moved into the templates. I know we tried to keep the templates as logic free as we can, but it is seem to be the best way to abstract out Gogo specific stuff.

sijis:
So, now that #123 is merged, It handles the instance size logic for a specific environment.

It looks like the only thing left ot address is forcing EC2 HC in dev: https://github.com/gogoair/foremast/blob/master/src/foremast/pipeline/construct_pipeline_block.py#L136

sijis:
Still present. https://github.com/gogoair/foremast/blob/master/src/foremast/pipeline/construct_pipeline_block.py#L228

PR #123, opened 2017-01-30
Stage asg

bazimov:
Set minimum default restriction for stage environment

### DIFF
--- a/src/foremast/pipeline/construct_pipeline_block.py
+++ b/src/foremast/pipeline/construct_pipeline_block.py
@@ -20,7 +20,7 @@ import json
 import logging
 from pprint import pformat
 
-from ..consts import DEFAULT_EC2_SECURITYGROUPS
+from ..consts import ASG_WHITELIST, DEFAULT_EC2_SECURITYGROUPS
 from ..utils import generate_encoded_user_data, get_template
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


sijis:
I'm really against this.
We should have a different mechanism to incorporate things like this, like a plugin.

jvasallo:
We need to remember that this library is a shared library by the world. We cannot impose limits on our end users (specifically over Spinnaker). Fully agree with @sijis on this one.

e4r7hbug:
Do you guys have an example that can help Bek go in the right direction?

e4r7hbug:
We've had some side discussions. Please change this to add `ASG_WHITELIST` to `data` that is passed into the Template. This will then enable template logic like:
```python
"min": {%- if data.app.appname not in data.ASG_WHITELIST %} 10 {% else %} {{ data.asg.min_inst }} {%- endif %},
```

e4r7hbug:
Will resolve #95.

e4r7hbug:
This still needs to pass `ASG_WHITELIST` into the `data` dictionary.

bazimov:
@e4r7hbug ASG_WHITELIST is in consts.py, not used in the construct_pipeline_block.py.

e4r7hbug:
Please refer to your diff.

ndcampbell:
I think this looks good to me. Have our internal templates been updated yet?

bazimov:
I am updating the internal templates soon.

PR #126, opened 2017-02-09
Added setup.cfg for bdist universal.

bazimov:


sijis:
👍 It'll help prevent a python specific version release :)

PR #127, opened 2017-02-09
fixed issue with wrongly named keywords

ndcampbell:


jvasallo:
Cool, nice find. This should fix that issue with multi-region deploys?

ndcampbell:
Should fix one of the issues at least.

PR #129, opened 2017-02-17
added cookie_ttl support and fixed bug with duplicate stickiness policy

ndcampbell:
As per requested, I added support for setting a LB cookie TTL for stickiness. This defaults to 0 which means no TTL.

I also expanded on the policy name to fix an issue with creating duplicate policies and not allowing a policy to change.

### DIFF
--- a/src/foremast/elb/create_elb.py
+++ b/src/foremast/elb/create_elb.py
@@ -165,14 +165,18 @@ class SpinnakerELB:
             if listener.get("stickiness"):
                 sticky_type = listener['stickiness']['type'].lower()
                 externalport = int(listener['loadbalancer'].split(":")[-1])
-                policyname = "{0}-{1}cookie-{2}".format(self.app, sticky_type, externalport)
                 if sticky_type == 'app':
+                    cookiename = listener['stickiness']['cookie_name']
+                    policyname = "{0}-{1}cookie-{2}-{3}".format(self.app, sticky_type, externalport, cookiename)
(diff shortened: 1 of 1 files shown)


src/foremast/elb/create_elb.py
```diff
@@ -165,14 +165,18 @@ def add_stickiness(self):
             if listener.get("stickiness"):
                 sticky_type = listener['stickiness']['type'].lower()
                 externalport = int(listener['loadbalancer'].split(":")[-1])
-                policyname = "{0}-{1}cookie-{2}".format(self.app, sticky_type, externalport)
                 if sticky_type == 'app':
+                    cookiename = listener['stickiness']['cookie_name']
+                    policyname = "{0}-{1}cookie-{2}-{3}".format(self.app, sticky_type, externalport, cookiename)
```
e4r7hbug: Save the format string in a new variable before the `if`.

ndcampbell: I did have it like that but now my policyname is different if ELB or APP sticky_type, so I was doing it in the conditional

e4r7hbug: The format string is the same. The arguments are what differ.
```
policyname_format = '{0}-{1}cookie-{2}-{3}'
policyname = policyname_format.format(self.app, sticky_type, externalport, cookiename)
```

ndcampbell: ohhh, I see what you mean. I hadnt thought about defining the string and then applying format later but that makes perfect sense. I'll update.

PR #133, opened 2017-02-22
added missing client for route53 delete func

bazimov:


e4r7hbug:
We'll take this for now, but we need some Test Cases.

PR #135, opened 2017-03-02
Feature/s3 pipeline

ndcampbell:
This adds core functionality for create an S3 bucket and attaching a policy. It does not handle any of the uploading, just basic s3 setup for s3/static content applications.


jvasallo:
This looks good to me, spoke about the create_s3 flag and its rename to create Archaius. I think long term we should see if we can leverage the deploy s3 app for Archaius.

PR #137, opened 2017-03-08
Feat/s3 spinnaker templates

sbasgall:
Added jinja templates for the new stages and the code to load them

### DIFF
--- a/src/foremast/pipeline/construct_pipeline_block_s3.py
+++ b/src/foremast/pipeline/construct_pipeline_block_s3.py
@@ -29,7 +29,6 @@ def construct_pipeline_block_s3(env='',
                              generated=None,
                              previous_env=None,
                              region='us-east-1',
-                             region_subnets=None,
                              settings=None,
                              pipeline_data=None):
     """Create the Pipeline JSON from template.
@@ -44,8 +43,6 @@ def construct_pipeline_block_s3(env='',
             Trigger.
         region (str): AWS Region to deploy to.
         settings (dict): Environment settings from configurations.
-        region_subnets (dict): Subnets for a Region, e.g.
-            {'us-west-2': ['us-west-2a', 'us-west-2b', 'us-west-2c']}.
 
     Returns:
         dict: Pipeline JSON template rendered with configurations.
--- a/src/foremast/pipeline/construct_pipeline_block_s3.py
+++ b/src/foremast/pipeline/construct_pipeline_block_s3.py
@@ -1,0 +1,85 @@
+#   Foremast - Pipeline Tooling
+#
+#   Copyright 2016 Gogo, LLC
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Construct a block section of Stages in a Spinnaker Pipeline."""
+import copy
+import json
+import logging
+from pprint import pformat
+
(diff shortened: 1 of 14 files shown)


ndcampbell:
We are also going to need an `s3-promote` stage. If the deployment strategy is redblack, it only deploys to a version and then you promote that into `LATEST`. I was thinking a flow like:

`s3-infra -> s3-deploy -> promotion-checkpoint -> s3-promote-latest -> $env checkpoint`

src/foremast/pipeline/construct_pipeline_block_s3.py
```diff
@@ -0,0 +1,84 @@
...
+from ..utils import generate_encoded_user_data, get_template
+
+LOG = logging.getLogger(__name__)
+
+
+def construct_pipeline_block_s3(env='',
+                             generated=None,
+                             previous_env=None,
+                             region='us-east-1',
+                             region_subnets=None,
```
ndcampbell: `region_subnets` does not seem to get used.

src/foremast/pipeline/create_pipeline_s3.py
```diff
@@ -0,0 +1,144 @@
...
+    """Manipulate Spinnaker Pipelines.
+
+    Args:
+        app (str): Application name.
+        trigger_job (str): Jenkins trigger job.
+        base (str): Base image name (i.e: fedora).
+        prop_path (str): Path to the raw.properties.json.
+    """
+
+    def render_wrapper(self, region='us-east-1'):
```
ndcampbell: We should abstract this out, it seems like we are duplicating code with the other pipeline types. I think it is fine for this PR but worth doing in the future.

src/foremast/pipeline/create_pipeline_s3.py
```diff
@@ -0,0 +1,144 @@
...
+        self.log.debug('Envs from pipeline.json: %s', pipeline_envs)
+
+        regions_envs = collections.defaultdict(list)
+        for env in pipeline_envs:
+            for region in self.settings[env]['regions']:
+                regions_envs[region].append(env)
+        self.log.info('Environments and Regions for Pipelines:\n%s',
+                      json.dumps(regions_envs, indent=4))
+
+        subnets = get_subnets()
```
ndcampbell: Do we need subnets for s3 deployments?

src/foremast/pipeline/create_pipeline_s3.py
```diff
@@ -0,0 +1,144 @@
...
+
+        pipelines = {}
+        for region, envs in regions_envs.items():
+            # TODO: Overrides for an environment no longer makes sense. Need to
+            # provide override for entire Region possibly.
+            pipelines[region] = self.render_wrapper(region=region)
+
+            previous_env = None
+            for env in envs:
```
ndcampbell: I dont think this block is needed since we dont need subnets

src/foremast/pipeline/create_pipeline_s3.py
```diff
@@ -0,0 +1,144 @@
...
+                except KeyError:
+                    self.log.info('%s is not available for %s.', env, region)
+                    continue
+
+                block = construct_pipeline_block_s3(
+                    env=env,
+                    generated=self.generated,
+                    previous_env=previous_env,
+                    region=region,
+                    region_subnets=region_subnets,
```
ndcampbell: No subnets needed for s3 deployments

sbasgall:
Subnet info is removed. 
When lambda support was added the realization occurred that there was a lot of duplicate code the pipeline module. I agree that going forward this should be streamlined.

ndcampbell:
Cool. Everything else looks good then other than my comment about needing a promote stage as well

src/foremast/templates/pipeline/pipeline_prod_s3.json.j2
```diff
@@ -0,0 +1,7 @@
+[
+    {% include "pipeline/stage-judgement-nonprod.json.j2" %},
+    {% include "pipeline/stage-infrastructure-setup-s3.json.j2" %},
+    {% include "pipeline/stage-deploy-s3.json.j2" %},
```
ndcampbell: Does this need the deploy_strategy logic too?

sbasgall: Yes

PR #138, opened 2017-03-08
added S3 website settings for making a bucket serve static content

ndcampbell:
Do you think website should be enabled or disabled by default?


jvasallo:
This looks fine to me, it is low risk and only touching new features. Nothing seems off, and everything seems accounted for. Thanks!

PR #139, opened 2017-03-13
Catches problems with empty or missing artifact path.

ndcampbell:
This fixes a bug where you could upload an empty directory and that would then sync to LATEST, basically removing LATEST.


jvasallo:
I thought this was a feature 😉 . Just kidding. Nice fix, and this could prevent some nasty user experiences. Thanks!

PR #140, opened 2017-03-14
Add SimpleDB IAM template

willkelleher:


sbasgall:
I tested this change locally and it works as expected.


One feature we may want to implement is a sdb resource name check to ensure it matches some combination of the project/repo name to isolate permissions.

ndcampbell:
This looks good to me 👍

PR #141, opened 2017-03-14
Add multipart upload permissions to s3 template

willkelleher:


ndcampbell:
I see no issues with this. Not sure if someone else wants to chime in.

PR #142, opened 2017-03-18
Support for modifying ASG Grace Period

jvasallo:
This will allow users to set their ASG grace period if their application takes longer to boot up than usual. We should consider splitting up into two variables that ultimately calculate hc_grace_period; os_grace_period and app_grace_period.

os_grace_period is usually best known by sysadmins (in regards to average boot time of base operating systems). Development teams would know how long it takes for their container to spin up. Ideally, os_grace_period should be hard coded and maintained by admins to the default time it takes to boot up a raw vanilla server. Devs should then be exposed to enabling only app_grace_period; since they see startup times in their logs.

os_grace_period: 180s
app_grace_period: 40s
hc_grace_period: 220s

### DIFF
--- a/src/foremast/pipeline/construct_pipeline_block.py
+++ b/src/foremast/pipeline/construct_pipeline_block.py
@@ -140,7 +140,8 @@ def construct_pipeline_block(env='',
     # Aggregate the default grace period, plus the exposed app_grace_period
     # to allow per repo extension of asg healthcheck grace period
     hc_grace_period = data['asg'].get('hc_grace_period')
-    hc_grace_period += data['asg'].get('app_grace_period')
+    app_grace_period = data['asg'].get('app_grace_period')
+    grace_period = hc_grace_period + app_grace_period
 
     data['app'].update({
         'appname': gen_app_name,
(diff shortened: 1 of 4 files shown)


sijis:
Changes look safe. 👍

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -137,6 +137,11 @@ def construct_pipeline_block(env='',
         data['asg'].update({'hc_type': 'EC2'})
         LOG.info('Switching health check type to: EC2')
 
+    # Aggregate the default grace period, plus the exposed app_grace_period
+    # to allow per repo extension of asg healthcheck grace period
+    hc_grace_period = data['asg'].get('hc_grace_period')
```
sijis: Why not make this explicit?
```
hc_grace_period = data['asg'].get('hc_grace_period') + data['asg'].get('app_grace_period')
```

jvasallo: That seems a bit too verbose for my tastes. How does it look now with separate variables?

sijis: Looks good :)

PR #143, opened 2017-04-01
Remove duplicate key

sijis:
Remove a duplicate key in the `pipeline.json.j2` template file.


jvasallo:
Yeah, that is an odd one. Thanks for fixing!

PR #148, opened 2017-04-19
Enabled custom SSH key pairs

jvasallo:
We had an unused app_ssh_key variable in our configs, I moved the key pair to the asg block seeing as that is where it is implemented. Renamed the variable to ssh_keypair.

### DIFF
--- a/_docs/configuration_files/application_json.rst
+++ b/_docs/configuration_files/application_json.rst
@@ -31,13 +31,6 @@ Describes the application.
 
     | *Default*: ``null``
 
-``app_ssh_key``
-*******************
-
-SSH key that your EC2 instances will use. Must already be created in AWS.
-
-    | *Default*: ``"{{ account }}_access"`` - {{ account }} being the AWS account in the configuration name
-
 ``eureka_enabled``
 ***********************
 
@@ -151,6 +144,13 @@ Minimum number of instances your auto-scaling group should have at all times. Th
 
     | *Type*: int
     | *Default*: ``1``
+
+``ssh_keypair``
+*******************
+
+SSH key that your EC2 instances will use. Must already be created in AWS.
+
+    | *Default*: ``"{{ account }}_{{ region }}_default"`` - {{ account }} being the AWS account in the configuration name
 
 ``subnet_purpose``
 ******************
--- a/_docs/configuration_files/application_json.rst
+++ b/_docs/configuration_files/application_json.rst
@@ -148,7 +148,7 @@ Minimum number of instances your auto-scaling group should have at all times. Th
 ``ssh_keypair``
 *******************
 
-SSH key that your EC2 instances will use. Must already be created in AWS.
+SSH key that your EC2 instances will use. Must already be created in AWS. This replaces the non-functional and deprecated app_ssh_key configuration key.
 
     | *Default*: ``"{{ account }}_{{ region }}_default"`` - {{ account }} being the AWS account in the configuration name
 
--- a/src/foremast/pipeline/construct_pipeline_block.py
+++ b/src/foremast/pipeline/construct_pipeline_block.py
@@ -143,6 +143,8 @@ def construct_pipeline_block(env='',
     app_grace_period = data['asg'].get('app_grace_period')
     grace_period = hc_grace_period + app_grace_period
     
+    # TODO: Migrate the naming logic to an external library to make it easier
+    #       to update in the future. Gogo-Utils looks like a good candidate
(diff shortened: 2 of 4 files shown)


src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -142,6 +142,12 @@ def construct_pipeline_block(env='',
     hc_grace_period = data['asg'].get('hc_grace_period')
     app_grace_period = data['asg'].get('app_grace_period')
     grace_period = hc_grace_period + app_grace_period
+    
+    ssh_keypair = data['asg'].get('ssh_keypair', None)
+    LOG.info('SSH keypair ({0}) used'.format(ssh_keypair))
+    if not ssh_keypair:
+        ssh_keypair = '{0}_{1}_default'.format(env, region)
```
sijis: The default should be defined in `configs.json`. It may avoid some of this logic.
Although, maybe we want the ability to *not* set a ssh key, for teams that do not want ssh access at all.

src/foremast/templates/configs/configs.json.j2
```diff
@@ -17,6 +16,7 @@
         "app_grace_period": 0,
         "max_inst": 3,
         "min_inst": 1,
+        "ssh_keypair": null,
```
sijis: The default value should be noted here. Since it appears that `app_ssh_key` is being refactored to this key (which is good), it should follow the same behaviour and default.

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -142,6 +142,12 @@ def construct_pipeline_block(env='',
     hc_grace_period = data['asg'].get('hc_grace_period')
     app_grace_period = data['asg'].get('app_grace_period')
     grace_period = hc_grace_period + app_grace_period
+    
+    ssh_keypair = data['asg'].get('ssh_keypair', None)
```
sijis: I'm thinking there should be a warning here about `app_ssh_key` no longer being used.

_docs/configuration_files/application_json.rst
```diff
@@ -152,6 +145,13 @@ Minimum number of instances your auto-scaling group should have at all times. Th
     | *Type*: int
     | *Default*: ``1``
 
+``ssh_keypair``
+*******************
+
+SSH key that your EC2 instances will use. Must already be created in AWS.
```
sijis: It should note that this replaced `app_ssh_key`.

_docs/configuration_files/application_json.rst
```diff
@@ -31,13 +31,6 @@ Describes the application.
 
     | *Default*: ``null``
 
-``app_ssh_key``
```
sijis: Maybe add deprecation notice?

src/foremast/pipeline/construct_pipeline_block.py
jvasallo: @sijis I confirmed, it was a 100% stale reference, whoever was using it, was definitely not using it anywhere. It basically was a stale key in the dictionary.

_docs/configuration_files/application_json.rst
jvasallo: See first comment.

_docs/configuration_files/application_json.rst
jvasallo: I will add this just to be clear, but state that it never worked. Migrate to this.

src/foremast/pipeline/construct_pipeline_block.py
jvasallo: We run into an issue with region. Can we sync up on this?

src/foremast/templates/configs/configs.json.j2
jvasallo: Region may not be populated, by moving that default logic to the code, I was able to handle the specified and unspecified case. Lets talk.

src/foremast/pipeline/construct_pipeline_block.py
jvasallo: Added a TODO to migrate this out but explained the issue that if we specify an temp value, region is not available when configs.json.j2 is parsed just yet, we talked about externalizing this bit to gogo-utils. Added a todo

src/foremast/templates/configs/configs.json.j2
sijis: OK. Region is populated later. Possibly moving this to gogo-utils may help longer term.

PR #151, opened 2017-04-19
Adding recursive cp before s3 sync

bazimov:
AWS does not have a feature for s3 bucket to bucket time based sync. We are placing cp command first then sync so our sync works as expected.


jvasallo:
@bazimov Did you try test this out to see how it looks?

bazimov:
@jvasallo yes tested, and works.

jvasallo:
This looks good to me. This is to solve the issue wheres s3 sync has issues with timestamps and exact size files. We needed to copy in place, then do sync to clean up stale files.

PR #152, opened 2017-04-20
Feature/shared s3

ndcampbell:
Added the ability to specify a shared bucket for S3 deployments. Looks like the following:

Master Bucket Config:

 "s3": {
     "shared_bucket_master": true
     "bucket_acl": "private",
     "bucket_policy": $policy,
     "website": {
         "enabled": true,
     }
  } 

NonMaster Shared Bucket Config:

  "s3": {
    "shared_bucket_target": "$sharedappname"
  } 

This creates a bucket prefixed with `common` and then the nonmaster deploys to a path with appname in the shared bucket.

### DIFF
--- a/src/foremast/exceptions.py
+++ b/src/foremast/exceptions.py
@@ -148,3 +148,7 @@ class PrimaryDNSRecordNotFound(ForemastError):
 class S3ArtifactNotFound(ForemastError):
     """Could not find Artifact to upload to S3"""
     pass
+
+class S3SharedBucketNotFound(ForemastError):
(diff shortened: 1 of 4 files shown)


jvasallo:
Looks good to me, either create docs or at least create an issue to create docs. Once that is done I am good to merge!

ndcampbell:
created #153 for the docs. We do not have docs for any s3 deployments yet so I think that should be added in a different PR

jvasallo:
Cool. Sounds like a plan!

https://github.com/gogoair/foremast/issues/153

PR #154, opened 2017-04-21
Fixed invalid app cookie stickiness policy names

jvasallo:
Some JS libraries ship with dots by default in their cookie name, this at least strips it out.

This will fix @willkelleher report of default cookies from Express.


sijis:
LGTM.

PR #155, opened 2017-04-29
feat: Add ability to use a dynamic config

sijis:
We are now able to load a python file 'config.py' as a source of
configuration details.

### DIFF
--- a/_docs/configuration_files/foremast_config.rst
+++ b/_docs/configuration_files/foremast_config.rst
@@ -1,8 +1,8 @@
 .. _foremast_config:
 
-============
-foremast.cfg
-============
+========================
+foremast.cfg / config.py
+========================
 
 .. contents::
    :local:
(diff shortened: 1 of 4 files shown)


sijis:
This add basic functionality and it essentially assumes that `config.py` is in the same directory that foremast is executed.

If this is eventually merged, I think we should also consider adding deprecation notice regarding `foremast.cfg` and eventually drop it in the next major release.

sijis:
Oh yeah, here's an example config.py
```
CONFIG = {

    'base': {
        'domain': 'example.com',
        'envs': 'dev,stage',
        'regions': 'us-east-1,us-west-2',
        'git_url': 'https://github.com',
        'gate_api_url': 'https://example.com:1234',
        'templates_path': '/path/to/github/foremast-templates',
        'default_ec2_securitygroups': 'my_office',
        'default_elb_securitygroups': 'all_apps',
        'gate_client_cert': '/path/to/my.pem',
        'gate_ca_bundle': '/path/to/bundle.pem',
        'types': 'ec2,lambda,manual',
    },
    'credentials': {
        'gitlab_token': 'yyyyyyyyyyyyyyyyyyyyyy',
        'slack_token': 'zzzzzzzzzzzzzzzzzzzzzzz',
    },

    'whitelists': {
        'asg_whitelist': 'myspecial_app',
    },

    'formats': {
        'domain': 'example.com',
    }
}
```

ndcampbell:
I really like this. It makes it so we can use Vault (or any source of secrets) for the tokens and SSL certs.

ndcampbell:
I see no issues with the code and I think it can be merged in since old static functionality is maintained.

PR #156, opened 2017-05-02
feat: Configurable timeouts for tasks

ckamradt:
Added [timeouts] section to config file.  The task running code checks to see if the currently running task is listed in the config and if so uses that timeout instead of the default.


ckamradt:
need to update the tests

ckamradt:
is it preferred to squash the test fix into the refactored commit?

sijis:
A separate commit is fine.

sijis:
You should rebase. Also, please update the relevant docs regarding these changes.

At quick glance I don't see the difference between TIMEOUT and  TASK_TIMEOUT. Maybe one of the variable may need renaming so it makes more sense, especially since it's exposed to the user/admin.

ndcampbell:
I think this looks good. Existing functionality seems to remain unchanged. Have you been able to test this locally and it works?

src/foremast/consts.py
```diff
@@ -155,7 +156,8 @@ def find_config():
                                  default='').split(','))
 GITLAB_TOKEN = validate_key_values(config, 'credentials', 'gitlab_token')
 SLACK_TOKEN = validate_key_values(config, 'credentials', 'slack_token')
-
+DEFAULT_TASK_TIMEOUT = validate_key_values(config, 'task_timeouts', 'default', default=120)
+TASK_TIMEOUTS = json.loads(validate_key_values(config, 'task_timeouts', 'envs', default="{}"))
```
sijis: Something seems strange around this.
Should default be `default={}` and not a quoted {}? Also, why is this wrapped in a `json.loads()`?

src/foremast/autoscaling_policy/create_policy.py
```diff
@@ -84,8 +84,7 @@ def prepare_policy_template(self, scaling_type, period_sec, server_group):
         self.log.info('Rendering Scaling Policy Template: {0}'.format(template_kwargs))
         rendered_template = get_template(template_file='infrastructure/autoscaling_policy.json.j2', **template_kwargs)
         print(rendered_template)
```
sijis: This should be some `LOG.xxx()` or removed.

src/foremast/utils/tasks.py
```diff
@@ -99,3 +98,48 @@ def check_task(taskid):
...
+        polls for task status.
+    """
+    wait = 2
+    max_attempts = int(timeout / wait)
+    return retry_call(partial(_check_task, taskid),
+                      max_attempts=max_attempts,
+                      wait=wait,
+                      exceptions=(AssertionError, ValueError))
+
+def wait_for_task(task_data):
```
sijis: I do like that this simplifies the existing model of `id = post_task(); check_task(id)`.

src/foremast/consts.py
ckamradt: the intent is to allow timeouts to differ per environment and per task.  ConfigParser requires that you convert any non-string values to the correct type.  eval'ing a dictionary seemed worse than loading json.  

the json.loads call is also why the {} is quoted:

```>>> json.loads({})
TypeError: the JSON object must be str, bytes or bytearray, not 'dict'
```

We could instead represent timeouts as a section per environment, something like:

```
[dev_task_timeouts]
Task1=240
Task2=480

[stage_task_timeouts]
Task1=500```

sijis: I would rather have the json than multiple sections.

sijis:
Its all good. 👍 
Nice job!

PR #157, opened 2017-05-04
Place log statements variable references after variables.

bazimov:
We had log statements referencing variables before they were set.


ndcampbell:
lgtm

PR #159, opened 2017-05-10
Initial concept for canary userdata in foremast

sbasgall:


ndcampbell:
I like this solution, I think it is the cleanest way to handle it. Do you need to update the `infrastructure/user_data.sh.j2` template to also include Canary?

sbasgall:
There is now a PR to modify that template in the public template repo
https://github.com/gogoair/foremast-template-examples/pull/2

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -153,6 +155,11 @@ def construct_pipeline_block(env='',
         'scalingpolicy': scalingpolicy,
     })
 
+    if data['app']['canary']:
```
ndcampbell: So as we discussed, I think moving everything under this one conditional is best. It keeps all the canary logic in one place

PR #160, opened 2017-05-10
fix: Re-enable linting

sijis:
Prospector has fixed the issues in their latest release.

Fix: https://github.com/landscapeio/prospector/pull/207
Release: https://pypi.python.org/pypi/prospector/0.12.5


ndcampbell:
lgtm

PR #161, opened 2017-05-11
Hotfix/fix remunerate

ndcampbell:
Fixed an issue where remunerate was not properly incrementing on multiple parallel/branch stages.

This fix now keeps track of a separate branch index  which is incremented if the previous generated stage was also a branch stage.

I also snuck in a fix for canary: false into the default templates


e4r7hbug:
Once new test cases pass, this is clear for merging.

PR #167, opened 2017-05-17
Cleanup Foremast Managed Lambda Policies

### LINKED ISSUE #165: Clean up removed Lambda triggers
If you remove a lambda trigger from your pipeline configs, it should be removed from Lambda. Currently, if removed or name is changed it does not clean anything up and can cause duplicate or abandoned triggers

ndcampbell:
added foremast-* to all SIDs and added functionality for cleaning up foremast created Lambda policies.

Before creating triggers, this deletes the policies for foremast-* SIDs. This will stop duplicate triggers from getting created and allow triggers to be removed from the configs. 

fixes #165

### DIFF
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -109,7 +109,7 @@ class APIGateway:
 
     def add_permission(self):
         """Add permission to Lambda for the API Trigger."""
-        statement_id = 'foremast-{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
+        statement_id = '{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
         principal = 'apigateway.amazonaws.com'
         lambda_alias_arn = get_lambda_alias_arn(self.app_name, self.env, self.region)
         add_lambda_permissions(function=lambda_alias_arn,
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -109,7 +109,7 @@ class APIGateway:
 
     def add_permission(self):
         """Add permission to Lambda for the API Trigger."""
-        statement_id = '{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
+        statement_id = 'foremast-{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
         principal = 'apigateway.amazonaws.com'
         lambda_alias_arn = get_lambda_alias_arn(self.app_name, self.env, self.region)
         add_lambda_permissions(function=lambda_alias_arn,
--- a/src/foremast/utils/awslambda.py
+++ b/src/foremast/utils/awslambda.py
@@ -143,7 +143,7 @@ def remove_all_lambda_permissions(app_name='', env='', region='us-east-1'):
     policy_json = json.loads(response['Policy'])
     LOG.debug("Found Policy: %s", response)
     for perm in policy_json['Statement']:
-        if perm['Sid'].startswith('foremast-'):
+        if perm['Sid'].startswith('foremast-') or app_name+'_' in perm['Sid']:
             lambda_client.remove_permission(FunctionName=lambda_arn,
                                             StatementId=perm['Sid'])
             LOG.info('removed permission: %s', perm['Sid'])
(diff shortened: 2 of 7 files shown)


sbasgall:
Do we want this to support qualifiers at this time? Should `remove_all_labda_permissions` iterate through all present lambda function qualifiers? For example, currently api_gateway permissions are not added to the non-qualified lambda function and will not be removed.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -109,7 +109,7 @@ def add_integration_response(self):
 
     def add_permission(self):
         """Add permission to Lambda for the API Trigger."""
-        statement_id = '{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
+        statement_id = 'foremast-{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
```
e4r7hbug: High chance of misspelling. Might be better to have a global constant.
```python
FOREMAST_PREFIX = 'foremast-'
```

src/foremast/utils/awslambda.py
```diff
@@ -118,3 +120,36 @@ def add_lambda_permissions(function='',
...
+            response = lambda_client.get_policy(FunctionName=arn)
+        except boto3.exceptions.botocore.exceptions.ClientError as error:
+            LOG.info("No policy exists for function %s, skipping deletion", arn)
+            LOG.debug(error)
+            return
+
+        policy_json = json.loads(response['Policy'])
+        LOG.debug("Found Policy: %s", response)
+        for perm in policy_json['Statement']:
+            if perm['Sid'].startswith('foremast-') or app_name+'_' in perm['Sid']:
```
e4r7hbug: I would like to see:
```python
legacy_sid_check = app_name + '_' in perm['Sid']
```
This would make the if statement read a little easier. Also, should it be `in`, or `.startswith()`?
```python
FOREMAST_PREFIX = 'foremast-'

legacy_prefix = app_name + '_'
if perm['Sid'].startswith(FOREMAST_PREFIX) or perm['Sid'].startswith(legacy_prefix):
```

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
ndcampbell: what if instead of a global prefix, I just add it to all SIDs on permission creation? They pass in the old statement_id and I prefix `foremast` to all indiscriminately?

src/foremast/utils/awslambda.py
ndcampbell: Should be startswith. I will update

ndcampbell:
@e4r7hbug how does that look now?

src/foremast/utils/awslambda.py
```diff
@@ -99,22 +100,59 @@ def add_lambda_permissions(function='',
     session = boto3.Session(profile_name=env, region_name=region)
     lambda_client = session.client('lambda')
     response_action = None
+    prefixed_sid = 'foremast-' + statement_id
```
e4r7hbug: Better, but there are still two `foremast-` string locations we have to maintain.

src/foremast/utils/awslambda.py
```diff
@@ -99,22 +100,59 @@ def add_lambda_permissions(function='',
...
+    lambda_alias_arn = get_lambda_alias_arn(app_name, env, region)
+    arns = (lambda_arn, lambda_alias_arn)
+
+    for arn in arns:
+        try:
+            response = lambda_client.get_policy(FunctionName=arn)
+        except boto3.exceptions.botocore.exceptions.ClientError as error:
+            LOG.info("No policy exists for function %s, skipping deletion", arn)
+            LOG.debug(error)
+            return
```
e4r7hbug: This could return immediately on the first ARN. Do we want that behaviour, or should we `continue` to the next ARN?

If we do want to return immediately, we should probably return `False` to indicate no changes. Then at the end return `True` to indicate changes made.

ndcampbell: Oh good catch

src/foremast/utils/awslambda.py
ndcampbell: Do you want it to be global to the package or to all of Foremast? Seems like if I make it global to all of Foremast I am over engineering this solution.

e4r7hbug: Your module level change looks good for me.

e4r7hbug:
I'm okay with these changes. I want to check with @sbasgall about his questions before merging.

ndcampbell:
I ended up addressing @sbasgall question. I dont iterate through all qualifiers, just the `$env` qualifier that we support.

e4r7hbug:
Got it. Alright, I do believe this is good to merge on Monday, unless you want to put in Test Cases.

PR #170, opened 2017-05-19
refactor: Package version

e4r7hbug:
Turns out, there was a hidden gem in https://setuptools.readthedocs.io/en/latest/setuptools.html#adding-support-for-revision-control-systems referring to a smart package, https://pypi.python.org/pypi/setuptools_scm.

```bash
$ foremast -v
3.17.2.dev10+gd62004be.d20170518
```

### DIFF
--- a/_docs/conf.py
+++ b/_docs/conf.py
@@ -24,7 +24,7 @@
 import shlex
 import subprocess
 
-from pkg_resources import packaging
+import pkg_resources
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


e4r7hbug:
Issue with `pip` installing `dev` releases over full releases. Working on way to make everything an `alpha` release.

e4r7hbug:
False alarm. `pip` behaves the way I want it to where `dev` versions are only installed when doing `pip install --pre foremast`. This will be going into `master` within the next week.

PR #173, opened 2017-05-23
Feature/data pipeline

ndcampbell:
added support for data pipeline deployments


src/foremast/datapipeline/__main__.py
```diff
@@ -0,0 +1,55 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""CLI entry point for aws datapipeline creation.
+
+Help: ``python -m src.foremast.awslambdaevent -h``
```
ckamradt: paste error?

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,96 @@
...
+                                pipelineObjects=pipelineObjects,
+                                parameterObjects=parameterObjects,
+                                parameterValues=parameterValues)
+        LOG.debug(response)
+        LOG.info("Successfully applied pipeline definition")
+
+    def get_pipeline_id(self):
+        """Finds the pipeline ID for configured pipeline"""
+
+        response = self.client.list_pipelines()
```
ckamradt: you're not guaranteed to get all of the results on the first call:  http://boto3.readthedocs.io/en/latest/reference/services/datapipeline.html#DataPipeline.Client.list_pipelines

src/foremast/pipeline/create_pipeline_datapipeline.py
```diff
@@ -0,0 +1,136 @@
...
+        for env in pipeline_envs:
+            for region in self.settings[env]['regions']:
+                regions_envs[region].append(env)
+        self.log.info('Environments and Regions for Pipelines:\n%s',
+                      json.dumps(regions_envs, indent=4))
+
+
+        pipelines = {}
+        for region, envs in regions_envs.items():
+            # TODO: Overrides for an environment no longer makes sense. Need to
```
ckamradt: should we file an issue to track?

tests/datapipeline/test_datapipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        assert False
+
+    properties['test_env']['datapipeline']['json_definition'] = BAD_DEF
+    mock_get_properties.return_value = properties
+    bad_dp = AWSDataPipeline(app='test_app', env='test_env', region='us-east-1', prop_path='other')
+    bad_dp.pipeline_id='1'
+    try:
+        bad_dp.set_pipeline_definition()
+        assert False
+    except awscli.customizations.datapipeline.translator.PipelineDefinitionError:
```
ckamradt: i believe you can use

`with assertRaises(awscli.customizations.datapipeline.translator.PipelineDefinitionError):
      bad_dp.set_pipeline_definition()`

tests/datapipeline/test_datapipeline.py
```diff
@@ -0,0 +1,96 @@
...
+    mock_boto3.return_value.create_pipeline.return_value = {}
+
+    dp = AWSDataPipeline(app='test_app', env='test_env', region='us-east-1', prop_path='other')
+    dp.create_datapipeline()
+    assert dp.pipeline_id == None
+
+@mock.patch('foremast.datapipeline.datapipeline.boto3.Session.client')
+@mock.patch('foremast.datapipeline.datapipeline.get_details')
+@mock.patch('foremast.datapipeline.datapipeline.get_properties')
+def test_set_pipeline_definition(mock_get_properties, mock_get_details, mock_boto3):
```
ckamradt: may be good to separate the happy path from the expected error into 2 separate test cases

src/foremast/datapipeline/datapipeline.py
ndcampbell: Ohh thanks, I will add paginating to it

src/foremast/datapipeline/__main__.py
ndcampbell: lol yup

src/foremast/pipeline/create_pipeline_datapipeline.py
ndcampbell: put in #177 for this

tests/datapipeline/test_datapipeline.py
ndcampbell: oh thanks, that looks better

tests/datapipeline/test_datapipeline.py
ndcampbell: Agreed, I will move it

src/foremast/templates/pipeline/pipeline_prod_datapipeline.json.j2
```diff
@@ -0,0 +1,5 @@
+[
+    {% include "pipeline/stage-judgement-nonprod.json.j2" %},
+    {% include "pipeline/stage-infrastructure-setup-datapipeline.json.j2" %}
+    {% endif %}
```
e4r7hbug: Wait, Jinja2 doesn't yell about this `{% endif %}` without a matching `{% if something %}`?

ndcampbell: oh, it does! I fixed it in our internal templates but forgot about this one. Good catch

tests/datapipeline/test_datapipeline.py
```diff
@@ -0,0 +1,101 @@
...
+@mock.patch('foremast.datapipeline.datapipeline.boto3.Session.client')
+@mock.patch('foremast.datapipeline.datapipeline.get_details')
+@mock.patch('foremast.datapipeline.datapipeline.get_properties')
+def test_create_datapipeline(mock_get_properties, mock_get_details, mock_boto3):
+    """Check data pipeline creation"""
+    generated = {"project": "test"}
+    properties = copy.deepcopy(TEST_PROPERTIES)
+    mock_get_details.return_value.data = generated
+    mock_get_properties.return_value = properties
+    mock_boto3.return_value.create_pipeline.return_value = {}
```
e4r7hbug: `create_pipeline()` should return what AWS would return, https://boto3.readthedocs.io/en/latest/reference/services/datapipeline.html#DataPipeline.Client.create_pipeline:
```python
{
    'pipelineId': 'string'
}
```
And we should assert that this was set properly as `AWSDataPipeline.pipeline_id`.

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,109 @@
...
+
+        pipelineObjects = translator.definition_to_api_objects(self.datapipeline_data['json_definition'])
+        parameterObjects = translator.definition_to_api_parameters(self.datapipeline_data['json_definition'])
+        parameterValues = translator.definition_to_parameter_values(self.datapipeline_data['json_definition'])
+        response = self.client.put_pipeline_definition(
+                                pipelineId=self.pipeline_id,
+                                pipelineObjects=pipelineObjects,
+                                parameterObjects=parameterObjects,
+                                parameterValues=parameterValues)
+        LOG.debug(response)
```
e4r7hbug: We should check `response['errored']` and raise a custom Exception or log a WARNING message. This would help prevent silent issues in the future.

src/foremast/pipeline/create_pipeline_datapipeline.py
```diff
@@ -0,0 +1,136 @@
...
+
+        pipeline_envs = self.environments
+        self.log.debug('Envs from pipeline.json: %s', pipeline_envs)
+
+        regions_envs = collections.defaultdict(list)
+        for env in pipeline_envs:
+            for region in self.settings[env]['regions']:
+                regions_envs[region].append(env)
+        self.log.info('Environments and Regions for Pipelines:\n%s',
+                      json.dumps(regions_envs, indent=4))
```
e4r7hbug: This is the fourth time this block of code has been copied and pasted. Make this into a method or utility.

src/foremast/pipeline/create_pipeline_datapipeline.py
```diff
@@ -0,0 +1,136 @@
...
+
+        wrapper = get_template(
+            template_file='pipeline/pipeline_wrapper.json.j2',
+            data=data)
+
+        return json.loads(wrapper)
+
+    def create_pipeline(self):
+        """Main wrapper for pipeline creation.
+        1. Runs clean_pipelines to clean up existing ones
```
ndcampbell: I agree, but I think that can be changed in my refactoring of the pipeline creation.

src/foremast/pipeline/create_pipeline_datapipeline.py
ndcampbell: I agree, I will address that in my refactoring of the pipeline creation code in a different PR.

src/foremast/datapipeline/__main__.py
```diff
@@ -0,0 +1,55 @@
...
+    add_env(parser)
+    add_properties(parser)
+    add_region(parser)
+    args = parser.parse_args()
+
+    logging.getLogger(__package__.split('.')[0]).setLevel(args.debug)
+
+    log.debug('Parsed arguments: %s', args)
+
+    datapipeline = AWSDataPipeline(app=args.app,
```
ndcampbell: What do you mean? This is how all the other modules have it

tests/datapipeline/test_datapipeline.py
```diff
@@ -0,0 +1,101 @@
...
+
+    dp = AWSDataPipeline(app='test_app', env='test_env', region='us-east-1', prop_path='other')
+    dp.create_datapipeline()
+    assert dp.pipeline_id == None
+
+@mock.patch('foremast.datapipeline.datapipeline.boto3.Session.client')
+@mock.patch('foremast.datapipeline.datapipeline.get_details')
+@mock.patch('foremast.datapipeline.datapipeline.get_properties')
+def test_good_set_pipeline_definition(mock_get_properties, mock_get_details, mock_boto3):
```
ndcampbell: yeah, I want to assert that an exception is NOT raised. I couldnt figure out how to do that. The test fails properly if the exception is raised

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,109 @@
...
+            else:
+                LOG.info("Checking for more pipelines")
+                response = self.client.list_pipelines(marker=marker)
+            all_pipelines.extend(response['pipelineIdList'])
+            hasMore = response['hasMoreResults']
+            marker = response.get('marker')
+            LOG.debug(response)
+        
+        for pipeline in all_pipelines:
+            if pipeline['name'] == self.datapipeline_data.get('name', self.app_name):
```
ndcampbell: would this really be any better? Currently I exit out early if found and I never need to relook it up. I think this would add complexity

ndcampbell:
@e4r7hbug I incorporated all of your feedback or left comments.

src/foremast/datapipeline/__main__.py
e4r7hbug: Yeah, but all the other Modules have drifted away from what we're doing in `foremast.runner`. We were discussing previously about getting rid of the `__main__.py` files because of that issue. If we can call `foremast.runner` like we do now, we can start to at least make the `__main__.py` files relevant in case someone is using them.

src/foremast/pipeline/create_pipeline_datapipeline.py
e4r7hbug: Why not do the work now so we can review it and know it's working in some capacity? The refactoring work later will then be to update the other Methods to use the new code.

tests/datapipeline/test_datapipeline.py
e4r7hbug: Assert the Function or Method was called. It would probably also be good to check that it was called with your expected arguments. If the Function or Method returned data, you could assert on that as well. This goes along with @sijis's methodology that everything should return something, a `True` at the bare minimum, to represent a successful call.

src/foremast/pipeline/create_pipeline_datapipeline.py
ndcampbell: Because I am afraid of increasing scope and time of this feature by refactoring. What is there currently works and is consistent with the other code and this PR is already long enough

src/foremast/datapipeline/__main__.py
ndcampbell: I guess I do not understand what you mean by making `__main__.py` use foremast.runner.

tests/datapipeline/test_datapipeline.py
ndcampbell: I added an assert here.

src/foremast/datapipeline/datapipeline.py
e4r7hbug: I think it would be better. We would get an O(1) lookup against a dictionary in pretty much all scenarios. The single `.get()` operation also tells me quickly that we either get the ID, or `None`, in one place without having to look at the `__init__()`.

If we also want to use Cyclomatic Complexity Number as an objective measure of complexity, the requested changes would reduce the CCN by one unit.

Before:
```bash
(foremast) ntangsurat@derigible ~/git/github.foremast feature/data_pipeline $ lizard src/foremast/datapipeline/datapipeline.py
================================================
  NLOC    CCN   token  PARAM  length  location
------------------------------------------------
      11      1    113      5      20 __init__@30-49@src/foremast/datapipeline/datapipeline.py
      14      1    101      1      16 create_datapipeline@51-66@src/foremast/datapipeline/datapipeline.py
      14      2     87      1      17 set_pipeline_definition@68-84@src/foremast/datapipeline/datapipeline.py
      11      4     87      1      14 get_pipeline_id@86-99@src/foremast/datapipeline/datapipeline.py
1 file analyzed.
==============================================================
NLOC    Avg.NLOC  AvgCCN  Avg.token  function_cnt    file
--------------------------------------------------------------
     57      12.5     2.0       97.0         4     src/foremast/datapipeline/datapipeline.py

=============================================================================================
No thresholds exceeded (cyclomatic_complexity > 15 or length > 1000 or parameter_count > 100)
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
        57      12.5     2.0       97.0        4            0      0.00    0.00
```

After:
```bash
(foremast) ntangsurat@derigible ~/git/github.foremast feature/data_pipeline $ lizard src/foremast/datapipeline/datapipeline.py
================================================
  NLOC    CCN   token  PARAM  length  location
------------------------------------------------
      11      1    113      5      20 __init__@30-49@src/foremast/datapipeline/datapipeline.py
      10      1    101      1      12 create_datapipeline@51-62@src/foremast/datapipeline/datapipeline.py
      14      2     87      1      17 set_pipeline_definition@64-80@src/foremast/datapipeline/datapipeline.py
       8      3     68      1      11 get_pipeline_id@82-92@src/foremast/datapipeline/datapipeline.py
1 file analyzed.
==============================================================
NLOC    Avg.NLOC  AvgCCN  Avg.token  function_cnt    file
--------------------------------------------------------------
     50      10.8     1.8       92.2         4     src/foremast/datapipeline/datapipeline.py

=============================================================================================
No thresholds exceeded (cyclomatic_complexity > 15 or length > 1000 or parameter_count > 100)
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
        50      10.8     1.8       92.2        4            0      0.00    0.00
```

Neither code blocks will handle the strange case where we have duplicate Data Pipeline names. Also, using `laboratory`, it looks like both current and proposed code runs in similar times with a small data set.

```bash
(foremast) ntangsurat@derigible ~/git/github.foremast feature/data_pipeline $ ./t.py
Current result: 0:00:00.170039
Proposed result:        0:00:00.158559
Current result: 0:00:00.166175
Proposed result:        0:00:00.152733
Current result: 0:00:00.189293
Proposed result:        0:00:00.195909
Current result: 0:00:00.186727
Proposed result:        0:00:00.175716
Current result: 0:00:00.171098
Proposed result:        0:00:00.167694
Current result: 0:00:00.171337
Proposed result:        0:00:00.180104
Current result: 0:00:00.183483
Proposed result:        0:00:00.174779
Current result: 0:00:00.188255
Proposed result:        0:00:00.177035
Current result: 0:00:00.177887
Proposed result:        0:00:00.170697
Current result: 0:00:00.173281
Proposed result:        0:00:00.168539
```

I'm willing to drop this issue.

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,114 @@
...
+                                pipelineId=self.pipeline_id,
+                                pipelineObjects=pipelineObjects,
+                                parameterObjects=parameterObjects,
+                                parameterValues=parameterValues)
+        except Exception as e:
+            LOG.warning(e)
+            raise ErrorCreatingDataPipeline
+        LOG.debug(response)
+        LOG.info("Successfully applied pipeline definition")
+        return True
```
e4r7hbug: Why not return `response`? It's already there and could be useful.

src/foremast/datapipeline/__main__.py
e4r7hbug:
```python
from foremast.runner import prepare_infrastructure

prepare_infrastructure()
```

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,114 @@
...
+        try:
+            pipelineObjects = translator.definition_to_api_objects(json_def)
+            parameterObjects = translator.definition_to_api_parameters(json_def)
+            parameterValues = translator.definition_to_parameter_values(json_def)
+            response = self.client.put_pipeline_definition(
+                                pipelineId=self.pipeline_id,
+                                pipelineObjects=pipelineObjects,
+                                parameterObjects=parameterObjects,
+                                parameterValues=parameterValues)
+        except Exception as e:
```
e4r7hbug: Don't catch a broad `Exception`. You were handling `awscli.customizations.datapipeline.translator.PipelineDefinitionError` before, so only handle that scenario. Otherwise, we will end up with masked failures.

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,114 @@
...
+                {"key": "app_name",
+                "value": self.app_name}
+                ]
+        try:
+            response = self.client.create_pipeline(name=self.datapipeline_data.get('name', self.app_name),
+                                               uniqueId=self.app_name,
+                                               description=self.datapipeline_data['description'],
+                                               tags=tags)
+            self.pipeline_id = response.get('pipelineId')
+        except Exception as e:
```
e4r7hbug: Don't catch a broad `Exception`. Only handle specific exceptions as needed. Otherwise, we will end up with masked failures.

src/foremast/pipeline/create_pipeline_datapipeline.py
e4r7hbug: Just drop this for now then.

src/foremast/pipeline/create_pipeline_datapipeline.py
e4r7hbug: Drop this, too, for now.

src/foremast/datapipeline/__main__.py
ndcampbell: oh ok, then I'll need to export everything as an environment variable.

ndcampbell:
Alright, @e4r7hbug addressed all your listed concerns

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,116 @@
...
+
+        session = boto3.Session(profile_name=self.env, region_name=self.region)
+        self.client = session.client('datapipeline')
+        self.pipeline_id = None
+
+    def create_datapipeline(self):
+        """Creates the data pipeline if it does not already exist
+
+        Returns:
+                json: the response of the Boto3 command
```
e4r7hbug: Are you sure this is a JSON string that is returned?

src/foremast/datapipeline/datapipeline.py
```diff
@@ -0,0 +1,116 @@
...
+
+        LOG.debug(response)
+        LOG.info("Successfully configured Data Pipeline - %s", self.app_name)
+        return response
+
+    def set_pipeline_definition(self):
+        """Translates the json definition and puts it on created pipeline
+
+        Returns:
+                json: the response of the Boto3 command
```
e4r7hbug: Are you sure this is a JSON string that is returned?

PR #176, opened 2017-05-30
added data pipeline config docs

ndcampbell:


jvasallo:
Did a small fix to wording, but it should be good! Merging after Tests.

PR #182, opened 2017-06-02
Move tagging to create lambda from update.

bazimov:
Misplaced tag option, now corrected.


ndcampbell:
lgtm, thanks for quick fix

PR #185, opened 2017-06-09
added paginating for finding lambda function arn

ndcampbell:


src/foremast/utils/awslambda.py
```diff
@@ -39,17 +39,23 @@ def get_lambda_arn(app, account, region):
...
+                lambda_arn = lambda_function['FunctionArn']
+                LOG.debug("Lambda ARN for lambda function %s is %s.", app, lambda_arn)
+                break
+
+    if not lambda_arn:
         LOG.fatal('Lambda function with name %s not found in %s %s', app, account, region)
-        raise LambdaFunctionDoesNotExist('Lambda function with name {0} not found in {1} {2}'.format(app, account,
-                                                                                                     region))
+        raise LambdaFunctionDoesNotExist('Lambda function with name {0} not found in {1} {2}'.format(app, account, region))
+    else:
```
e4r7hbug: Remove `else`.

src/foremast/utils/awslambda.py
```diff
@@ -39,17 +39,23 @@ def get_lambda_arn(app, account, region):
...
-        if lambda_function['FunctionName'] == app:
-            lambda_arn = lambda_function['FunctionArn']
-            LOG.debug("Lambda ARN for lambda function %s is %s.", app, lambda_arn)
-            return lambda_arn
-    else:
+    lambda_arn = None
+    paginator = lambda_client.get_paginator('list_functions')
+
+    for lambda_functions in paginator.paginate():
+        if lambda_arn:
```
e4r7hbug: Small readability improvement would move `if: break` below the big `for`.

src/foremast/utils/awslambda.py
```diff
@@ -39,17 +39,23 @@ def get_lambda_arn(app, account, region):
...
+            break
+        for lambda_function in lambda_functions['Functions']:
+            if lambda_function['FunctionName'] == app:
+                lambda_arn = lambda_function['FunctionArn']
+                LOG.debug("Lambda ARN for lambda function %s is %s.", app, lambda_arn)
+                break
```
e4r7hbug: We discussed about how to remove the double `break`. This second level `for` loop would need to be refactored out to a new function, but we're okay with keeping for now.

e4r7hbug:
Let's go with the changes we discussed, but we should try this using https://boto3.readthedocs.io/en/latest/reference/services/lambda.html#Lambda.Client.get_function to remove all the manual looping in general.

PR #188, opened 2017-06-21
fix: Check Task should use a specific exception

e4r7hbug:
`foremast.utils.tasks.check_task()` needs to provide a specific exception when a Spinnaker Task fails to complete in time.


e4r7hbug:
@ckamradt, you should also check my code.

sijis:
I see no issues with this. Unless I hear otherwise, I'll merge in Monday.

jvasallo:
Monday has come and gone...and come and gone. Merging in! 👍

PR #190, opened 2017-06-22
Added passing of source ARN to permissions in lambda function policy …

sbasgall:
…and add function polices to alias and unqualified functions

This works if the apigateway resource matches the lambda function name.

There is no cleanup yet for a subsequent update of an already created apigateway resource.


sbasgall:
As this stands with the initial commit if the apigateway resource exists already ( such as a redeploy that updates the resource ) boto errors with:

```
botocore.errorfactory.ConflictException: An error occurred (ConflictException) when calling the CreateResource operation: Another resource with the same parent already has this name: gogologapipoc
```

We will probably have to delete the resource and recreate it if it exists and is being changed. This could lead to a small blip in the service during deployment.

This PR adds the correct triggers to both the aliases and unqualified functions which seems to be required to run the aliased function from an apigateway trigger.

sbasgall:
apigateway aws api does have an `update-resource` method which should solve the blip.
We need to change the `create_resource` function to `create_or_update_resource` and include the proper logic.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -202,19 +231,21 @@ def create_api(self):
...
-                                                       pathPart=resource_name)
-        resource_id = created_resource['id']
-        self.log.info("Successfully created resource")
+        try:
+            created_resource = self.client.create_resource(restApiId=self.api_id,
+                                                           parentId=parent_id,
+                                                           pathPart=resource_name)
+            resource_id = created_resource['id']
+            self.log.info("Successfully created resource")
+        except:
```
e4r7hbug: O, that's why this is failing. Put at least a `pass` for the `except`.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -112,12 +112,41 @@ def add_permission(self):
         statement_id = '{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
         principal = 'apigateway.amazonaws.com'
         lambda_alias_arn = get_lambda_alias_arn(self.app_name, self.env, self.region)
+        lambda_arn = get_lambda_arn(self.app_name, self.env, self.region)
+        resource_name = self.trigger_settings.get('resource', '')
+        resource_name = resource_name.replace("/", "")
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/{}/{}/{}'.format(self.region, self.account_id, self.api_id, self.env, self.trigger_settings['method'], resource_name)
```
e4r7hbug: What Resource is this ARN for? Use a more descriptive variable name.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -112,12 +112,41 @@ def add_permission(self):
...
+                               source_arn=source_arn)
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/*/*/{}'.format(self.region, self.account_id, self.api_id, resource_name)
+        add_lambda_permissions(function=lambda_alias_arn,
+                               statement_id=statement_id,
+                               action='lambda:InvokeFunction',
+                               principal=principal,
+                               env=self.env,
+                               region=self.region,
+                               source_arn=source_arn)
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/{}/{}/{}'.format(self.region, self.account_id, self.api_id, self.env, self.trigger_settings['method'], resource_name)
```
e4r7hbug: This is the same as the first `source_arn` assignment. By using a unique variable name, this duplication can be prevented.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -112,12 +112,41 @@ def add_permission(self):
         statement_id = '{}_api_{}'.format(self.app_name, self.trigger_settings['api_name'])
         principal = 'apigateway.amazonaws.com'
         lambda_alias_arn = get_lambda_alias_arn(self.app_name, self.env, self.region)
+        lambda_arn = get_lambda_arn(self.app_name, self.env, self.region)
+        resource_name = self.trigger_settings.get('resource', '')
+        resource_name = resource_name.replace("/", "")
```
e4r7hbug: Inconsistent quote usage. Please stick to single quotes like the line above this.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -112,12 +112,41 @@ def add_permission(self):
...
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/{}/{}/{}'.format(self.region, self.account_id, self.api_id, self.env, self.trigger_settings['method'], resource_name)
         add_lambda_permissions(function=lambda_alias_arn,
+                               statement_id=statement_id + self.trigger_settings['method'],
+                               action='lambda:InvokeFunction',
+                               principal=principal,
+                               env=self.env,
+                               region=self.region,
+                               source_arn=source_arn)
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/*/*/{}'.format(self.region, self.account_id, self.api_id, resource_name)
```
e4r7hbug: Use a more descriptive variable name.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -112,12 +112,41 @@ def add_permission(self):
...
+                               source_arn=source_arn)
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/{}/{}/{}'.format(self.region, self.account_id, self.api_id, self.env, self.trigger_settings['method'], resource_name)
+        add_lambda_permissions(function=lambda_arn,
+                               statement_id=statement_id + self.trigger_settings['method'],
+                               action='lambda:InvokeFunction',
+                               principal=principal,
+                               env=self.env,
+                               region=self.region,
+                               source_arn=source_arn)
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/*/*/{}'.format(self.region, self.account_id, self.api_id, resource_name)
```
e4r7hbug: This is a duplicate of the second `source_arn`.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -112,12 +112,41 @@ def add_permission(self):
...
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/{}/{}/{}'.format(self.region, self.account_id, self.api_id, self.env, self.trigger_settings['method'], resource_name)
         add_lambda_permissions(function=lambda_alias_arn,
+                               statement_id=statement_id + self.trigger_settings['method'],
+                               action='lambda:InvokeFunction',
+                               principal=principal,
+                               env=self.env,
+                               region=self.region,
+                               source_arn=source_arn)
+        source_arn = 'arn:aws:execute-api:{}:{}:{}/*/*/{}'.format(self.region, self.account_id, self.api_id, resource_name)
+        add_lambda_permissions(function=lambda_alias_arn,
```
e4r7hbug: The only difference between the first two `add_lambda_permissions()` calls is `function=lambda_alias_arn` and `function=lambda_arn`. Those values can be used in a `for` loop.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -202,19 +231,21 @@ def create_api(self):
         self.log.info("Successfully created API")
         return api_id
 
-    def create_resource(self, parent_id=""):
+    def create_or_update_resource(self, parent_id=""):
```
e4r7hbug: There is at least one call to the method `create_resource()` that needs to be updated to the new method name.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -209,13 +233,15 @@ def create_resource(self, parent_id=""):
...
-                                                       pathPart=resource_name)
-        resource_id = created_resource['id']
-        self.log.info("Successfully created resource")
-        return resource_id
+        resource_name = resource_name.replace('/', '')
+        if not self.resource_id:
+            created_resource = self.client.create_resource(restApiId=self.api_id,
+                                                           parentId=parent_id,
+                                                           pathPart=resource_name)
+            self.resource_id = created_resource['id']
```
sbasgall: I don't like to do this here but its worth moving create_resource call out of find_resource_id function.

When we initialize on a fresh deployment self.resource_id will be None but it is needed to pass to attach_method. Would moving the attach method call from the end of this file into this create_resource function be less bad?

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -236,6 +262,8 @@ def attach_method(self, resource_id):
 
     def setup_lambda_api(self):
         """A wrapper for all the steps needed to setup the integration."""
+        self.create_resource(self.parent_id)
+        self.attach_method(self.resource_id)
```
sbasgall: Should this be moved into the create_resource function?

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
sbasgall: This try/except was unintended to be in the commit. Bouncing back and forth with production issues.

sbasgall:
This PR is to fix a non functional API Gateway trigger for lamba functions.

This fix is limited to functions that use API Gateway with some specific assumptions -

1. Only one resource is declared in the function json configuration

2. The parent of this resource must be `/` and declared as such in the json configuration

3. For this single resource only one HTTP method is designated

4. Proxy configuration is not supported

This will meet a great number of users needs and we can determine how we want to support multiple resources, multiple methods, resources with parents that aren't `/`. Proxy support should be easier and implementation can go forward without addressing the first 3 requirements.

sbasgall:
Also, until it is decided how to handle multiple resources, resources with a parent that isn't `/` and multiple methods in a single API Gateway trigger. I propose we delay adding the logic to update these parameters. As implemented if resource settings are changed in the function configuration json it will require a manual deletion of the existing resource to apply.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
e4r7hbug: I like having the `attach_method()` call here in case someone wants to override this Method and it's clearer to read. Passing a class attribute as an argument seems a little strange, but no harm.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
e4r7hbug: I would say keeping them separate is a good idea.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -79,13 +79,7 @@ def find_resource_id(self):
...
                 resource_id = resource['id']
                 self.log.info("Found Resource ID for: %s", resource['path'])
-                self.attach_method(resource_id)
-                break
-        else:
-            resource_id = self.create_resource(parent_id=parent_id)
-            self.attach_method(resource_id)
-
-        return resource_id
+        return resource_id, parent_id
```
e4r7hbug: This is fine, but I would like to see a follow up Pull Request. We can define the class attributes in `__init__()`. Then this method should assign the class attributes directly.

I have no issues with returning the values. Sijis and I like when functions and methods return at least something (makes testing easier). I would suggest using a `collections.namedtuple` in the future.

e4r7hbug:
To be clear, I think the overview in https://github.com/gogoair/foremast/pull/190#issuecomment-310534974 is an acceptable way forward in development.

PR #192, opened 2017-06-25
feat: on function update also update tags

sbasgall:
Currently we only set tags on lambda `create_function`. This adds tag updates to `update_function_configuration`.

If tags exist it will overwrite them without complaint. This lets us skip a check for the existence of these tag keys.

This code wouldn't be needed for people only using lambda deployments after creation tagging was added but it will ensure they remain correct if they are manually changed or removed. 

The documentation for boto's `update_function_configuration` does not include a Tags dict like boto's `create_function` does. Although it's possible this is supported and not documented since Lambda tags are relatively new.

This PR is based off the documented capabilities of boto.

### DIFF
--- a/src/foremast/awslambda/awslambda.py
+++ b/src/foremast/awslambda/awslambda.py
@@ -195,7 +195,7 @@ class LambdaFunction(object):
 
         lambda_arn = get_lambda_arn(self.app_name, self.env, self.region)
         self.lambda_client.tag_resource(
-            Resource = lambda_arn,
+            Resource=lambda_arn,
             Tags={
                 'app_group': self.group,
                 'app_name': self.app_name
--- a/src/foremast/awslambda/awslambda.py
+++ b/src/foremast/awslambda/awslambda.py
@@ -21,7 +21,8 @@ import boto3
 from tryagain import retries
 
 from ..exceptions import RequiredKeyNotFound
-from ..utils import get_details, get_properties, get_role_arn, get_security_group_id, get_subnets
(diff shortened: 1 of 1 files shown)


src/foremast/awslambda/awslambda.py
```diff
@@ -21,7 +21,8 @@
 from tryagain import retries
 
 from ..exceptions import RequiredKeyNotFound
-from ..utils import get_details, get_properties, get_role_arn, get_security_group_id, get_subnets
+from ..utils import (get_details, get_lambda_arn, get_properties, get_role_arn,
```
sbasgall: Is this formatting update preferred over the original style?

e4r7hbug: I would let `isort` manage the import line formatting for consistency. Eventually it'll wrap with parenthesis when it gets too long.

src/foremast/awslambda/awslambda.py
```diff
@@ -190,6 +191,15 @@ def update_function_configuration(self, vpc_config):
                 raise SystemExit(message)
 
             raise
+        LOG.info('Updating Lambda function tags')
+
+        lambda_arn = get_lambda_arn(self.app_name, self.env, self.region)
+        self.lambda_client.tag_resource(
+            Resource = lambda_arn,
```
e4r7hbug: Spaces around equals visually makes this look like an assignment.

e4r7hbug:
That's annoying that `tag_resource()` takes an ARN while everything else can use the Function name. Everything thing looks good to me. Clear for merging.

Issue #153, opened 2017-04-20
Docs for S3 deployments


ndcampbell:

jvasallo:
@ndcampbell Can we close this with: https://github.com/gogoair/foremast/pull/228

jvasallo:
This was resolved.

PR #194, opened 2017-06-28
Enhancement to s3 canary deployments

ndcampbell:
Added a feature so that when `deploy_strategy=canary`, an S3 bucket will deploy to a `CANARY` path as well. 

The versioned path still remains the source of deployment and `CANARY` and `LATEST` get synced from there.


src/foremast/s3/s3deploy.py
```diff
@@ -68,8 +68,10 @@ def setup_pathing(self):
         s3_format = "s3://{}"
         version_path = path_format.format(self.bucket, self.s3path, self.version).replace('//', '/')
```
sijis: Somehow, this continous `.replace('//', '/')` seems strange. Maybe there's a way to do this earlier or replace with a function?

ndcampbell: Yeah, I can refactor this so that a function is doing it

ndcampbell:
@sijis added small doc on this deploy_strategy. We have issue #153 for creating docs on S3 deployments but we do not have anything on that yet. I also renamed those variables

sijis:
All good.

PR #180, opened 2017-06-01
Refactor/pipeline creation

ndcampbell:
Refactored pipeline creation to have a lot less copy/paste code

PR #195, opened 2017-06-29
fix errors in onetime pipelines for ec2 deployments

ndcampbell:
This fixes onetime pipelines for ec2 deployments. Merging in #180 should get close to enabling this for other pipeline types. 

I had to add logic for handling Pipeline ID as this changed in spinnaker since we last used the onetime functionality.

### DIFF
--- a/src/foremast/pipeline/create_pipeline.py
+++ b/src/foremast/pipeline/create_pipeline.py
@@ -186,12 +186,12 @@ class SpinnakerPipeline:
         pipeline_id = None
         found = False
         for pipeline in pipelines:
-            right_app_and_region = (pipeline['application'] == self.app_name) and (region in pipeline['name']) 
+            correct_app_and_region = (pipeline['application'] == self.app_name) and (region in pipeline['name']) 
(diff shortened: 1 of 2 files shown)


jvasallo:
Still reviewing, but this is a good fix for the one time deploys. Good fix. Thanks!

src/foremast/pipeline/create_pipeline.py
```diff
@@ -172,19 +172,29 @@ def get_existing_pipelines(self):
...
 
         Returns:
             str: pipeline_id if existing, empty string of not.
         """
         pipelines = self.get_existing_pipelines()
         pipeline_id = None
+        found = False
         for pipeline in pipelines:
-            if (pipeline['application'] == self.app_name) and (region in pipeline['name']):
+            right_app_and_region = (pipeline['application'] == self.app_name) and (region in pipeline['name'])
```
jvasallo: Any opportunity to make this a bit more verbose such as `correct` or `valid`? The right threw me off a bit. :)

PR #196, opened 2017-06-29
S3 Tagging

sbasgall:
S3 tagging feature


ndcampbell:
This all looks good to me. It should be tested first as I want to make sure it does not do anything weird with new and existing buckets

ndcampbell:
LGTM. merge after tests complete

PR #197, opened 2017-06-29
Refactor Construct Pipeline Block to Segregate ec2 logic

ndcampbell:
This is an attempt to break #180 into smaller PRs

This just separates out ec2 logic from create_pipeline and construct_pipeline_block. This makes it easier to then refactor so that the separate classes for s3 and lambda can be removed. Makes the calls more generic

### DIFF
--- a/src/foremast/pipeline/construct_pipeline_block.py
+++ b/src/foremast/pipeline/construct_pipeline_block.py
@@ -69,14 +69,38 @@ def check_provider_healthcheck(settings, default_provider='Discovery'):
 
     return ProviderHealthCheck(providers=health_check_providers, has_healthcheck=has_healthcheck)
 
+def get_template_name(env, pipeline_type):
+    """Generates the correct template name based on pipeline type
+
+    Args:
+        env (str): environment to generate templates for
+        pipeline_type (str): Type of pipeline like ec2 or lambda
+    
+    Returns:
+        str: Name of template
+    """
+    if pipeline_type == 'ec2':
+        if env.startswith('prod'):
+            template_name = 'pipeline/pipeline_{}.json.j2'.format(env)
+        else:
+            template_name = 'pipeline/pipeline_stages.json.j2'
+    else:
+        if env.startswith('prod'):
+            template_name = 'pipeline/pipeline_{}_{}.json.j2'.format(env, pipeline_type)
+        else:
+            template_name = 'pipeline/pipeline_stages_{}.json.j2'.format(pipeline_type)
+
+    return template_name
 
 def construct_pipeline_block(env='',
                              generated=None,
                              previous_env=None,
                              region='us-east-1',
-                             region_subnets=None,
                              settings=None,
-                             pipeline_data=None):
+                             pipeline_data=None,
+                             pipeline_type='ec2',
+                             **kwargs):
+
     """Create the Pipeline JSON from template.
 
     This handles the common repeatable patterns in a pipeline, such as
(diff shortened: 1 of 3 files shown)


src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -70,13 +70,38 @@ def check_provider_healthcheck(settings, default_provider='Discovery'):
...
+    Args:
+        env (str): environment to generate templates for
+        pipeline_type (str): Type of pipeline like ec2 or lambda
+    
+    Returns:
+        str: Name of template
+    """
+    if pipeline_type == 'ec2':
+        if env.startswith('prod'):
+            template_name = 'pipeline/pipeline_{}.json.j2'.format(env)
```
e4r7hbug: It might be better to separate the string formats from the `.format()` method call. This would also require the use of keyword style format strings.
```python
if:
    template_name_format = '{env}'

template_name = template_name_format.format(env=env)
```
I think this looks a little cleaner and is slightly more extensible. It also might help us see how to collapse this `if` stack.

ndcampbell: @e4r7hbug ok, good idea. I updated this and collapsed the conditional into just one level.

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -100,21 +126,62 @@ def construct_pipeline_block(env='',
...
+    })
+
+    LOG.debug('Block data:\n%s', pformat(data))
+
+    template_name = get_template_name(env, pipeline_type)
+    pipeline_json = get_template(template_file=template_name, data=data)
+    return pipeline_json
+
+
+def ec2_pipeline_setup(**kwargs):
```
e4r7hbug: Is there a reason why we're only using `**kwargs`? It would be nice to break this out into individual keyword arguments so we can avoid the assignments.

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -100,21 +126,62 @@ def construct_pipeline_block(env='',
...
-        template_name = 'pipeline/pipeline_stages.json.j2'
 
+    LOG.info('%s block for [%s].', env, region)
     LOG.debug('%s info:\n%s', env, pformat(settings))
 
     gen_app_name = generated.app_name()
-    user_data = generate_encoded_user_data(env=env, region=region, app_name=gen_app_name, group_name=generated.project)
+    data = copy.deepcopy(settings)
+
+    setup_kwargs = {
```
e4r7hbug: This variable isn't used outside the `if`. Should tuck it into there.

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -100,21 +126,62 @@ def construct_pipeline_block(env='',
...
+        'appname': gen_app_name,
+        'settings': settings,
+        'env': env,
+        'region': region,
+        'project': generated.project
+    }
+
+    if pipeline_type == 'ec2':
+        setup_kwargs['region_subnets'] = kwargs['region_subnets']
+        data = ec2_pipeline_setup(**setup_kwargs)
```
e4r7hbug: If we're not reusing or modifying the keyword arguments, we should pass in keyword arguments instead of unpacking the `dict`.

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -70,13 +70,39 @@ def check_provider_healthcheck(settings, default_provider='Discovery'):
...
+
+
 def construct_pipeline_block(env='',
                              generated=None,
                              previous_env=None,
                              region='us-east-1',
-                             region_subnets=None,
                              settings=None,
-                             pipeline_data=None):
+                             pipeline_data=None,
```
e4r7hbug: Missing docstring changes for new keyword arguments.

src/foremast/pipeline/construct_pipeline_block.py
ndcampbell: Nope, no good reason. I had originally had this cause I think ti was going to be a ton of args but it is only 5. I'll break them out

src/foremast/pipeline/construct_pipeline_block.py
ndcampbell: agreed, i'll break these apart

src/foremast/pipeline/construct_pipeline_block.py
ndcampbell: added this

src/foremast/pipeline/create_pipeline.py
```diff
@@ -233,23 +213,27 @@ def create_pipeline(self):
...
+                pipeline_block_data = {
+                    "env": env,
+                    "generated": self.generated,
+                    "previous_env": previous_env,
+                    "region": region,
+                    "settings": self.settings[env],
+                    "pipeline_data": self.settings['pipeline'],
+                }
+
+                if self.settings['pipeline']['type'] == 'ec2':
```
e4r7hbug: We should get a test case to compare old and new behaviour.

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -93,28 +119,72 @@ def construct_pipeline_block(env='',
...
+    gen_app_name = generated.app_name()
+
+    if pipeline_type == 'ec2':
+        data = ec2_pipeline_setup(
+            appname=gen_app_name,
+            settings=settings,
+            env=env,
+            region=region,
+            project=generated.project,
+            region_subnets=kwargs.get('region_subnets'))
```
ndcampbell: Yeah, you are right. I expected to have more but that never happened

src/foremast/pipeline/create_pipeline.py
ndcampbell: added test

e4r7hbug:
Thanks for the changes. I think this looks good for merging after the conflicts have been resolved. You might want to resolve some style linting issues if you get the chance.

PR #203, opened 2017-07-13
Add account level security group configuration options

sijis:
These changes allow to specific security groups for elb and ec2 instances per account.

Currently the only option is to set sg for all account. This gives a little more flexibility on what sg are added to a specific account.

### DIFF
--- a/src/foremast/consts.py
+++ b/src/foremast/consts.py
@@ -26,6 +26,7 @@ descending order. First found wins.
 .. literalinclude:: ../../src/foremast/templates/configs/foremast.cfg.example
    :language: ini
 """
+import ast
 import json
 import logging
 import sys
@@ -148,6 +149,18 @@ def _remove_empty_entries(entries):
     return entries
 
 
+def _convert_string_to_native(value):
+    """Convert a string to its native python type"""
+    result = None
+
+    try:
+        result = ast.literal_eval(value)
+    except (SyntaxError, ValueError):
+        # Likely a string
+        result = value.split(',')
+    return result
+
+
 def _generate_security_groups(config_key):
     """Read config file and generate security group dict by environemnt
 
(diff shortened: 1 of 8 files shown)


src/foremast/consts.py
```diff
@@ -139,6 +141,53 @@ def find_config():
...
+            entries.pop(count)
+    return entries
+
+
+def _convert_string_to_native(value):
+    """Convert a string to its native python type"""
+    result = None
+
+    try:
+        result = ast.literal_eval(value)
```
e4r7hbug: I don't know if we need `ast` here. If we're possibly expecting JSON, maybe we should use `json`? A test case for this sole function would help clarify what the incoming data is.

src/foremast/consts.py
```diff
@@ -139,6 +141,53 @@ def find_config():
...
+
+
+def _generate_security_groups(config_key):
+    """Read config file and generate security group dict by environemnt
+
+    Args:
+        config_key (str): Configuration file key
+
+    Returns:
+        dict: of environments in {$env: [$group1, group2]} format
```
e4r7hbug: Probably don't need the `$` in the example, looks fine without.

src/foremast/consts.py
```diff
@@ -139,6 +141,53 @@ def find_config():
...
+    try:
+        result = ast.literal_eval(value)
+    except (SyntaxError, ValueError):
+        # Likely a string
+        result = value.split(',')
+    return result
+
+
+def _generate_security_groups(config_key):
+    """Read config file and generate security group dict by environemnt
```
e4r7hbug: Spell check and period.

src/foremast/consts.py
sijis: It could be a dict too, not just json.

It needs to handle a `config.py` too.

e4r7hbug: Bwah.

src/foremast/consts.py
```diff
@@ -140,6 +142,54 @@ def find_config():
     return configurations
 
 
+def _remove_empty_entries(entries):
+    """Remove empty entries in a list"""
+    valid_entries = []
+    for entry in entries:
```
e4r7hbug: Minor optimization would be `set(entries)`. Not a real concern, but something to note.

e4r7hbug:
Looks good for me.

src/foremast/consts.py
sijis: addressed

PR #204, opened 2017-07-15
Keep old s3 tags when new tags added.

bazimov:
There is a logic to add app_group, app_name tag. S3 api does put tagset which overwrites any other tags that exist in s3 bucket. This logic keeps old tags as well addition to app_group, app_name.

### DIFF
--- a/src/foremast/s3/s3apps.py
+++ b/src/foremast/s3/s3apps.py
@@ -22,9 +22,10 @@ from botocore.client import ClientError
 
 from ..exceptions import S3SharedBucketNotFound
 from ..utils import (get_details, get_dns_zone_ids, get_properties,
-                     update_dns_zone_record)
+                     update_dns_zone_record, generate_s3_tags)
 
 LOG = logging.getLogger(__name__)
+
 
 class S3Apps(object):
     """Handles infrastructure around depolying static content to S3. Setups Bucket and Policy"""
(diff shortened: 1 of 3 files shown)


src/foremast/s3/s3apps.py
```diff
@@ -116,14 +116,37 @@ def _set_bucket_dns(self):
...
+        """Regular put_bucket_tagging sets TagSet which overwrites old tags.
+        Below logic keeps the old tags in place as well.
+        """
+        app_group_tag = {'Key': 'app_group', 'Value': self.group}
+        app_name_tag = {'Key': 'app_name', 'Value': self.app_name}
+        name_tag_count, group_tag_count = 0, 0
+        try:
+            # Get current tags list, if no tags exist will get an exception.
+            result = self.s3client.get_bucket_tagging(Bucket=self.bucket)['TagSet']
+            for item in result:
```
sijis: Wouldn't this be simpler to blindly copy all tags found an reinsert them?
The gathering of the tags should be a separate method too.

bazimov: @sijis Every key value is different dictionary in the list. So, I cannot update the value of a dictionary as is. I have to loop each dictionary and update if there is app_group, app_name. If they are not updated  (count = 0) then I know it did not have app_group, app_name before so we provide later below.

sijis: If you are in fact getting every tag, then the code should look something like this:
```
for item in result:
    item.update({'Key': item['Key'], 'Value': item['Value']})
```

You have a few conditionals for 'app_group' and 'app_name', which makes me think you are filtering the tags.

src/foremast/utils/generate_s3_tags.py
```diff
@@ -0,0 +1,9 @@
+def generated_tag_data(tags):
+    """Gets normal dictionary then converts to S3 tag set list."""
```
sijis: Update docstring to include something about the `tags` parameter.

sijis:
One last thing and it'll be good for merge.

PR #205, opened 2017-07-19
fix: Create S3 lambda trigger SID with hash of rules

willkelleher:
This prevents an issue where multiple lambda triggers for the same bucket end up with conflicting SIDs.

### DIFF
--- a/src/foremast/awslambda/s3_event/s3_event.py
+++ b/src/foremast/awslambda/s3_event/s3_event.py
@@ -15,6 +15,7 @@
 #   limitations under the License.
 
 import json
+import hashlib
 import logging
 
 import boto3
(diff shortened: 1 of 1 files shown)


sijis:
Could there be a possibility of orphaned triggers when changing the format?

willkelleher:
@sijis Only if there is already the possibility of orphaning a tag when it is removed from the rules set, I think.  I haven't looked closely but I though you delete all triggers and reapply the new set each time.

sijis:
Finally reviewed this.
I don't currently see any code that does cleanups for s3 events, so there will be orphaning, however, it could simply be added as we prefix all `StatementId` with `foremast-`.

The changes in question will solve the issue, so I'll merge this in knowing orphaning will occur.

PR #208, opened 2017-07-26
tests: Sanity check IAM Policy templates

e4r7hbug:
We need a way to automatically validate any IAM Policy templates will render to JSON properly. We can come back to this in the future and move the logic to a `foremast validate templates iam` command or something.


sijis:
@e4r7hbug and I have been discussing offline on how to make the error of the the file with the syntax error more prominent once it fails. Currently the file that has a syntax issue is noted during the test running sequences not the error output. It just makes it hard to easily determine which file had the issue.

sijis:
All good now.

PR #209, opened 2017-07-27
Refactor/move ec2 bake data

ndcampbell:
This is a continuation of #197. It was also part of #180 which was closed for being too large of a PR.

This abstracts out the ec2 bake logic from the render_wrapper, so that single function can be used for other pipeline types.


sijis:
@ndcampbell Is this still being worked on?

ndcampbell:
@sijis yeah, definitely still want this. It isnt the highest priority, but it is part of enabling me to remove all the duplicate code we have around pipeline construction. I took it out of WIP

PR #210, opened 2017-07-28
create new list to avoid mutation errors

ndcampbell:
This was causing security groups to be appended multiple times and anything in pervious stages would impact later a stages. 

Create a new list each time.


sijis:
Use sorted() so it's easy to see the output when debugging

ndcampbell:
so like `list(DEFAULT_EC2_SECURITYGROUPS[env]).sorted()`?

e4r7hbug:
Other way around. Sorted actually returns a list, so you only need `sorted([])`.

ndcampbell:
Ok, changed

PR #211, opened 2017-07-28
fix: Allow access to DynamoDB table/name/stream/

e4r7hbug:
For Lambda Functions to use DynamoDB as a Trigger, the IAM Role needs to have access to the `table/name/stream/` directory:
```
There was an error creating the trigger: Cannot access stream arn:aws:dynamodb:us-east-1:222572804561:table/name/stream/2017-07-28T18:53:40.930. Please ensure the role can perform the GetRecords, GetShardIterator, DescribeStream, and ListStreams Actions on your stream in IAM.
```


jvasallo:
Small permission change, looks good.

PR #212, opened 2017-08-03
Add tagging to security groups

bazimov:
Minor enhancement adding tags to the security groups.

### DIFF
--- a/src/foremast/securitygroup/create_securitygroup.py
+++ b/src/foremast/securitygroup/create_securitygroup.py
@@ -121,11 +121,8 @@ class SpinnakerSecurityGroup(object):
         self.log.debug('SG reference rules: %s', non_cidr)
         return non_cidr, cidr
 
-    def add_tags(self, group_id):
-        """
-        Add tags to security group.
-        Args:
-            group_id (str): Security group id.
+    def add_tags(self):
+        """Add tags to security group.
 
         Returns:
             True: Upon successful completion.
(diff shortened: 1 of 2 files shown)


src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -118,6 +121,40 @@ def _process_rules(self, rules):
         self.log.debug('SG reference rules: %s', non_cidr)
         return non_cidr, cidr
 
+    def add_tags(self, group_id):
+        """
+        Add tags to security group.
```
ndcampbell: Follow the format of the other doc strings, move this to be right after the `"""` and line between description and args

src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -162,10 +199,13 @@ def add_cidr_rules(self, rules):
                 if 'InvalidPermission.Duplicate' in str(error):
                     self.log.debug('Duplicate rule exist, that is OK.')
                 else:
-                    msg = 'Unable to add cidr rules to {}'.format(rules['app'])
+                    msg = 'Unable to add cidr rules to {}'.format(rule.get('app'))
                     self.log.error(msg)
                     raise SpinnakerSecurityGroupError(msg)
 
+        # Tag security group
+        self.add_tags(group_id)
```
ndcampbell: Is `add_cidr_rules` the right place to call the function? Seems like it might make more sense under `create_security_group` and do it right after `add_cidr_rules`.

bazimov: I thought so too, I just placed there because group_id was already called. I moved the group_id call into function as well then placed in `create_security_group` which is correct place yes.

src/foremast/securitygroup/create_securitygroup.py
bazimov: Fixed.

PR #213, opened 2017-08-08
Linting!

sijis:
This is an attempt to lint all the code.

The goal is at the end break CI if linting fails.


sijis:
@e4r7hbug it's all done! :)

e4r7hbug:
Awesome! This is going in before anything else can get in the way.

PR #216, opened 2017-08-17
Add application specific links in instance view

ViRaL95:
@ndcampbell


src/foremast/app/create_app.py
```diff
@@ -19,10 +19,9 @@
 """
 import logging
 from pprint import pformat
-
```
ndcampbell: This line should still be there.

src/foremast/app/create_app.py
```diff
@@ -87,11 +86,26 @@ def create_app(self, pipeline_configs=None):
...
+            pipeline_configs (dict): A dictionary containing the pipline.json info
+
+        Returns:
+            pipeline_configs (dict): The dictionary now containing instance links
+        """
+        if pipeline_configs['type'] == 'ec2':
+            for key, value in LINKS.items():
+                if key not in pipeline_configs['instance_links']:
+                    pipeline_configs['instance_links'][key] = value
+        return pipeline_configs
```
ndcampbell: I am iffy on this logic. I dont like that we are mutating a dictionary like this out of band. You do not actually use the return here. 

I think it might make more sense to make a copy of the dictionary and return that. Or to make the pipeline_configs a class attribute and modify `self.configs` without passing it around to functions

jvasallo:
Hey @ViRaL95 Can you add a bit of info around this PR in a comment or original body? Looks like you are adding Link support?

src/foremast/consts.py
```diff
@@ -210,6 +210,7 @@ def _generate_security_groups(config_key):
 APP_FORMATS = extract_formats(config)
 GATE_CLIENT_CERT = expandvars(expanduser(validate_key_values(config, 'base', 'gate_client_cert', default='')))
 GATE_CA_BUNDLE = expandvars(expanduser(validate_key_values(config, 'base', 'gate_ca_bundle', default='')))
+LINKS = _convert_string_to_native(validate_key_values(config, 'urls', 'spinnaker_links', {}))
```
ndcampbell: it might make more sense for the group to be named something like `default_links` instead of `urls`. We have other URLs elsewhere in the config.

src/foremast/templates/infrastructure/app_data.json.j2
```diff
@@ -24,10 +24,15 @@
                 {% if not loop.last %},{% endif %}
                 {% endfor %}
                 ]
-
-            }
+            },
+            "instanceLinks": [{
+                "title": "Internal Tools", "links":
```
ndcampbell: I know @sijis called it Internal Tools, but I think `Instance Links` or just `Links` would be more clear to developers

src/foremast/app/create_app.py
ViRaL95: good catch

ViRaL95:
This is a merge request to include Spinnaker Links on Spinnaker. That way if a user included documentation links, health check links, or tailon links it would include this on Spinnaker

ViRaL95:
@ndcampbell Check out my Changes !! 💃

src/foremast/app/create_app.py
```diff
@@ -85,13 +86,22 @@ def create_app(self, pipeline_configs=None):
         Raises:
             AssertionError: Application creation failed.
         """
+        self.pipeline_configs = pipeline_configs
```
sijis: I think this should be moved into `__init__()`. I find it strange to create a new property in some random method. I think it should be passed in when you instantiate `SpinnakerApp` class.

src/foremast/app/create_app.py
```diff
@@ -85,13 +86,22 @@ def create_app(self, pipeline_configs=None):
...
         jsondata = get_template(template_file='infrastructure/app_data.json.j2',
-                                appinfo=self.appinfo, pipeline_configs=pipeline_configs)
+                                appinfo=self.appinfo, pipeline_configs=self.pipeline_configs)
 
         wait_for_task(jsondata)
 
         self.log.info("Successfully created %s application", self.appname)
         return
+
+    def add_to_instance_links(self):
```
sijis: I'm always cautious about inline mutations of objects. I would rather prefer this return the calculation/appending of links.

This may need some debug/info output. I'd probably suggest `debug` before mutation and `info` after mutation.

src/foremast/app/create_app.py
```diff
@@ -39,13 +38,15 @@ class SpinnakerApp:
     Attributes:
         appinfo (dict): A dictionary containing the provided arguments
         appname (str): The name of the application.
+        pipeline_configs (dict): The dictionary containing the info from pipeline.json
```
ViRaL95: sure

src/foremast/app/create_app.py
```diff
@@ -85,13 +86,22 @@ def create_app(self, pipeline_configs=None):
...
+                                appinfo=self.appinfo, pipeline_configs=self.pipeline_configs)
 
         wait_for_task(jsondata)
 
         self.log.info("Successfully created %s application", self.appname)
         return
+
+    def add_to_instance_links(self):
+        """Appends on existing instance links"""
+        if self.pipeline_configs['type'] == 'ec2':
```
ViRaL95: There are no cluster's view for any application other than those that are ec2. So we would not have a Links window for us :). This if statement isn't necessary because issues  shouldnt arise if we added the links even if the app isnt of ec2 type. Lets see what @ndcampbell thinks

src/foremast/consts.py
```diff
@@ -210,6 +210,7 @@ def _generate_security_groups(config_key):
 APP_FORMATS = extract_formats(config)
 GATE_CLIENT_CERT = expandvars(expanduser(validate_key_values(config, 'base', 'gate_client_cert', default='')))
 GATE_CA_BUNDLE = expandvars(expanduser(validate_key_values(config, 'base', 'gate_ca_bundle', default='')))
+LINKS = _convert_string_to_native(validate_key_values(config, 'default_links', 'spinnaker_links', {}))
```
ViRaL95: True, this is a spinnaker tool and naming it something like spinnaker_links is a bit redundant

src/foremast/app/create_app.py
ViRaL95: Now that I think about it, this if statement shouldnt be there at all. That way if Spinnaker does allow a cluster view for lambda types we wouldnt have to remove this line of code to account for that. Ill remove it

ViRaL95:
Hey so I kind of mucked the rebase up so it may seem kind of weird. Thats why there are so many commits. But anyways this merge request essentially moves the retrieval of templates to a separate method (retrieve_template) and then returns all the instance_links in a dictionary vs mutating pipeline_configs directly. I also wrote two unit tests.  @ndcampbell

requirements.txt
```diff
@@ -5,6 +5,7 @@ jinja2
 murl
 pyapi-gitlab
 requests
+requests-mock
```
ndcampbell: This should go in `requirements-dev.txt` since it is a testing requirement

src/foremast/app/create_app.py
ndcampbell: This like should still be there :p

tests/app/test_create_app.py
```diff
@@ -0,0 +1,76 @@
...
+    instance_links = spinnaker_app.retrieve_instance_links()
+    assert instance_links == {"test1": "https://test1.com"}, "Instance Links are not being retrieved properly"
+
+    pipeline_config = { "instance_links": {} }
+    spinnaker_app = SpinnakerApp(pipeline_config=pipeline_config)
+    instance_links = spinnaker_app.retrieve_instance_links()
+    assert instance_links == {"test1": "https://test1.com", "test2": "https://test2.com"}, "Instance Links are not being retrieved properly"
+
+
+@mock.patch('foremast.app.create_app.API_URL', new='https://spinnaker.build.gogoair.com:7777')
```
ndcampbell: This should not default to a real URL. Unittests should work without internet access and we dont want to show our internal URLs on the open source code.

tests/app/test_create_app.py
```diff
@@ -0,0 +1,76 @@
...
+    app = "UnitTestApp"
+    email = "UnitTestApp@gogoair.com"
+    project = "UnitTest"
+    repo = "UnitTest"
+
+    accounts = [{
+        "type": "aws",
+        "name": "dev",
+    }]
+    mock_api_url = 'https://spinnaker.build.gogoair.com:7777/credentials'
```
ndcampbell: Nope, change this to something like `example.com`. nothing related to our internal URLs

tests/app/test_create_app.py
```diff
@@ -0,0 +1,76 @@
...
+    """Checks if instance_links is being populated properly"""
+    pipeline_config = {
+        "instance_links": {
+            "test3": "https://test2.com",
+            "test4": "https://test4.com",
+        }
+    }
+    spinnaker_app = SpinnakerApp(pipeline_config=pipeline_config)
+    instance_links = spinnaker_app.retrieve_instance_links()
+    assert instance_links == {"test1": "https://test1.com"}, "Instance Links are not being retrieved properly"
```
ndcampbell: I am not sure I understand this test. I would expected instance_links to equal `test3` and `test4` right? Why are you asserting that they equal `test`?

ViRaL95: this is because this first test is simply testing the retrieve_instance_links method which returns a dictionary containing all items not in pipeline_config.

requirements.txt
ViRaL95: nice catch

tests/app/test_create_app.py
ViRaL95: The returned value is then added into pipeline_configs

tests/app/test_create_app.py
ViRaL95: sure

ViRaL95:
@ndcampbell unitt tests passed

tests/app/test_create_app.py
ndcampbell: If that is the case, shouldnt it return `test2` as well?

ViRaL95: Nope so this method simply returns all the items in LINKS and not in pipeline_configs. This dictionary is then added into pipeline_configs to provide a dictionary that contains only unique items

ndcampbell: But you set LINKS to `'foremast.app.create_app.LINKS', new={"test1": "https://test1.com", "test2": "https://test2.com"}`. So why is `test1` and `test2` not returned?

ViRaL95: because "https://test2.com" is in pipeline_configs['instance_links']. I test to see if two values are the same in either dictionary.

ndcampbell: Where are you seeing `test2` being in `pipeline_configs['instance_links']`? I see test3 and test4.

ndcampbell: OHHH the url. You should make this more clear. Part of a unit test is to help understand how the code works

ndcampbell: and what is the string `"Instance Links are not being retrieved properly"` at the end of the assert doing?

ViRaL95: Sorry Ill include better comments for the unit tests. The string afterwards explains why the two values are not equal. Essentially why the unit test did not pass

ViRaL95: okay looks like i fixed that

PR #217, opened 2017-08-22
fix: explicitly encode lambda trigger rules

willkelleher:
This worked as-is with Python 2.7 but apparently I didn't test it with Python 3.


ndcampbell:
Does this still work for Python 2.7?

willkelleher:
should still work in 2.7, yes

PR #218, opened 2017-08-23
Preventing setting of username in RDS-DB.

jvasallo:
This patch forces App Name to be used for the user name to prevent folks from specifying any user.


jvasallo:
Thanks @sbasgall !

PR #219, opened 2017-08-23
fix: changed logic to accept blank filter pattern

sbasgall:


jvasallo:
This looks like a good fix, you are handling the case where you receive an empty string. I see no problem with this fix. Thanks!

PR #221, opened 2017-08-28
Feature: Security Group self reference in config.py

sijis:
This allows `config.py` to add self referencing application entries to security groups.

In `config.py`, you have:
```
'default_securitygroup_rules': {
    'bastion': [{'start_port': '22', 'end_port': '22', 'protocol': 'tcp'}],
    'serviceapp': [{'start_port': '8080', 'end_port': '8080', 'protocol': 'tcp' }],
    '$self': [{'start_port': '9000', 'end_port': '9000', 'protocol': 'tcp' }],
},
```
The generated result is (that would be posted to Spinnaker):
```
$ python -m foremast.securitygroup -a myapp -p /tmp/test.json -r "us-east-1" -e "dev" -d
....
2017-08-28 14:29:39,280 [DEBUG] foremast.utils.templates:get_template:83 - Rendered JSON:
{
    "application": "myapp",
    "description": "Create Security Group: myapp",
    "job": [
        {
            "credentials": "dev",
            "description": "something useful.",
            "detail": "",
            "name": "myapp",
            "regions": [
                "us-east-1"
            ],
            "securityGroupIngress": [
                
                {
                    "name": "bastion",
                    "type": "tcp",
                    "startPort": 22,
                    "endPort": 22
                    
                },
                
                {
                    "name": "serviceapp",
                    "type": "tcp",
                    "startPort": 8080,
                    "endPort": 8080
                    
                },
                
                {
                    "name": "myapp",
                    "type": "tcp",
                    "startPort": 9000,
                    "endPort": 9000
                    
                }
                
            ],
            "subnet": "none",
            "type": "upsertSecurityGroup",
            "user": "foremast",
            "vpcId": "vpc-xxxxx"
        }
    ]
}

```
Since we do not know the application beforehand, we use the keyword `$self` that updates the reference with the correct application name during the security group calculation phase.


ndcampbell:
This looks good to me. I dont have any specific feedback, seems useful and necessary.

Just make sure we get docs :)

src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -194,11 +194,21 @@ def add_cidr_rules(self, rules):
 
         return True
 
+    def resolve_self_references(self, rules):
+        """Resolves `$self` references to actual application name in security group rules."""
+        resolved_rules = {}
+        for app, rule in rules.items():
```
e4r7hbug: I think this would be easier with a `rule = dict.get(); if rule:`.

tests/test_securitygroup.py
```diff
@@ -81,14 +81,68 @@ def test_misconfiguration(get_details, get_properties):
...
+                'start_port': '22',
+                'end_port': '22',
+                'protocol': 'tcp'
+            },
+        ]
+    }
+
+    sg = SpinnakerSecurityGroup(app='myapp')
+    ingress = sg.resolve_self_references(test_sg)
+    assert 'myapp' in ingress
```
e4r7hbug: Probably good to check the list of ports as well. Would hate to have that slip through by accident in the future.

src/foremast/securitygroup/create_securitygroup.py
sijis: Are you suggesting something like this:
```
for app in rules:
  if app == '$self':
    app = self.app_name
  resolved_rules[app] = rules.get(app)
```
or
```
rule = rules.pop('$self')
if rule:
   rules[self.app_name] = rule
return rules
```

e4r7hbug: Second one. Might as well use the `dict` features.

tests/test_securitygroup.py
sijis: Updated test to check port.

PR #222, opened 2017-08-29
Fix issue creating security groups

sijis:
If we have self-referencing rules (aka `$self`) defined in `config.py`, then new apps will fail as there is a dependency loop going on. We can't create the sg since there is a rule referencing itself.

This changes the behavior on how we create security groups.

*New apps*
First create an sg, if none found
then update the sg with the new rules

*Existing apps*
Update sg with new rules

### DIFF
--- a/src/foremast/securitygroup/create_securitygroup.py
+++ b/src/foremast/securitygroup/create_securitygroup.py
@@ -209,8 +209,29 @@ class SpinnakerSecurityGroup(object):
         resolved_ingress = self.resolve_self_references(ingress)
         return resolved_ingress
 
-    def create_security_group(self):
+    def _create_security_group(self, ingress):
         """Send a POST to spinnaker to create a new security group.
+
+        Returns:
+            boolean: True if created successfully
+
+        """
+        template_kwargs = {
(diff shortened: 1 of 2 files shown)


src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -209,19 +209,46 @@ def update_default_rules(self):
...
+            boolean: True if created successfully
+
         Raises:
             ForemastConfigurationFileError: Missing environment configuration or
                 misconfigured Security Group definition.
         """
         ingress_rules = []
 
         try:
+            _ = get_security_group_id(name=self.app_name, env=self.env, region=self.region)
```
e4r7hbug: Name the variable instead of making it look like we throw it away.

src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -209,19 +209,46 @@ def update_default_rules(self):
...
+            self.log.debug('Security Group ID %s found for %s.', _, self.app_name)
```
e4r7hbug: Put this in an `else` clause.

PR #225, opened 2017-09-06
fix: multiple lambda s3 trigger event creation

willkelleher:
AWS operations need to be applied to the bucket once for all events that use that bucket.


sbasgall:
Did a session with @willkelleher and we fixed this PR to work as expected. LGTM - Do we want to require tests for this section of code that was improved? There are none to date.

sijis:
@sbasgall It would makes sense to do tests cases. (It doesn't need to be 100% coverage.)
It would be better to do them now as the code is fresh in everyone's mind. Going back and doing it later will be much harder, I think.

willkelleher:
@sijis @sbasgall added a couple tests

PR #226, opened 2017-09-06
added sorted to make a copy of the list so global variable is not mutated

ndcampbell:


sijis:
We are good with this.

PR #228, opened 2017-09-10
S3 Pipeline Documentation

jvasallo:
We have announced S3 support externally a few times now, but without docs our users would not be able to leverage it. I would consider this the official release of S3 pipelines to our OSS community. Great work team!


ndcampbell:
Looks good to me. I didnt see any typos and data looks accurate

PR #230, opened 2017-09-14
Feature/alpha s3

ndcampbell:
adds alpha pathing for a 3-part deployment to prod. Alpah -> Canary -> Latest

### DIFF
--- a/src/foremast/s3/s3deploy.py
+++ b/src/foremast/s3/s3deploy.py
@@ -107,6 +107,8 @@ class S3Deployment(object):
         """
         if promote_stage.lower() == 'alpha':
             self._sync_to_uri(self.s3_canary_uri)
+        elif promote_stage.lower() == 'canary':
+            self._sync_to_uri(self.s3_latest_uri)
         else:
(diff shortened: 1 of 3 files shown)


src/foremast/s3/s3deploy.py
```diff
@@ -88,20 +89,26 @@ def _path_formatter(self, suffix):
...
-        """Promote artifact version to LATEST."""
-        self._sync_to_uri(self.s3_latest_uri)
+    def promote_artifacts(self, promote_stage='latest'):
+        """Promote artifact version to dest.
+
+        Args:
+            promote_stage (string): Stage that is being promoted
+        """
+        if promote_stage.lower() == 'alpha':
+            self._sync_to_uri(self.s3_canary_uri)
```
jvasallo: Could we make it a bit explicit by also adding 'canary' and what happens? Its safe to assume today that if not 'alpha' it should go to LATEST today, but long term I can see this becoming a bug if we or anyone adds more stages.

jvasallo: Thanks!

PR #234, opened 2017-09-22
Remove security group restriction

sijis:
This removes that restriction when checking the cidr in a security group.

### DIFF
--- a/src/foremast/securitygroup/create_securitygroup.py
+++ b/src/foremast/securitygroup/create_securitygroup.py
@@ -74,8 +74,7 @@ class SpinnakerSecurityGroup(object):
         generated = get_details(app=self.app_name)
         self.group = generated.data['project']
 
-    @staticmethod
-    def _validate_cidr(rule):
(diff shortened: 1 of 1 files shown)


src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -90,9 +89,7 @@ def _validate_cidr(rule):
         except (ipaddress.NetmaskValueError, ValueError) as error:
             raise SpinnakerSecurityGroupCreationFailed(error)
 
-        if network.prefixlen < 13:
-            msg = 'The network range ({0}) specified is too open.'.format(rule['app'])
-            raise SpinnakerSecurityGroupCreationFailed(msg)
```
bazimov: Should we log warning instead of removing this if statement?

jvasallo: I wouldn't want any issues in the OSS release as companies may have different requirements. Good fix.

PR #237, opened 2017-10-18
Clean-up: Example S3 Bucket Name in Templates

jvasallo:
It appears the s3 template seems to have a reference so some example bucket. I can't find a reason for it to be there so I aligned it with the dynamodb.json template with the built in Jinja2 last feature.


e4r7hbug:
Yes, this fix appears to be the correct way of doing the S3 Template. Please update the Test Cases to reflect the correct behaviour.

jvasallo:
@e4r7hbug Sorry for the delay this should be fixed.

e4r7hbug:
Looks good now. Thanks!

PR #240, opened 2017-10-26
fix: Improve template retrieval errors

e4r7hbug:
Document and clean up template retrieval errors.


sijis:
I see not issues with this.

PR #241, opened 2017-10-26
fix: Return Task status, update docstrings

e4r7hbug:
We should return something useful from checking a Task status. The most useful thing we have is the status result from the poller. Also, updating some old docstrings with more information.


sijis:
Not seeing an issue here.

PR #242, opened 2017-10-26
refactor: Get Subnets

e4r7hbug:
Need to simplify getting list of Subnets from Spinnaker and update documentation.


src/foremast/utils/subnets.py
```diff
@@ -42,11 +76,19 @@ def get_subnets(
         region (str): AWS Region to find Subnets for.
 
     Returns:
```
sijis: I think this refactored function should just return 1 'format'.
Maybe
```
{'account': 'x', 'region': ['az']}
```

Or maybe this is split into two functions:
`get_all_subnets()`, `get_subnet(account=x)`

src/foremast/utils/subnets.py
```diff
@@ -26,6 +26,40 @@
 
 LOG = logging.getLogger(__name__)
 
+SubnetAttributes = namedtuple('SubnetAttributes', [
```
sijis: I haven't looked all the details as closely, but I like where this is headed.

src/foremast/utils/subnets.py
e4r7hbug: Yes, I noticed that TODO. Will work that in somehow.

PR #247, opened 2017-12-04
removed hardcoded image size limit

ndcampbell:
This limit can be enforced in templates if necessary. It was put in as a cost saving measure but should not be hard coded in.


sijis:
I don't see an issue with this.
We should remove these sort of things when we find them.

The alternative is to add the check in the jinja template.

sijis:
@ndcampbell I'll leave it up to you to cut a new release.

ndcampbell:
Thanks. I cut v3.28.1

PR #248, opened 2017-12-13
Added support for removing duplicate security group references based on names

jvasallo:
We have a case where we want to replace a previously required Security Group with a new security group. However, this caused two security groups to be attached when we really only needed one. This finds a security group based on name, and if both the original and replacement exist, remove the original and just use the replacement security group.


src/foremast/elb/create_elb.py
```diff
@@ -82,6 +82,11 @@ def make_elb_json(self):
         security_groups.append(self.app)
         security_groups.extend(self.properties['security_group']['elb_extras'])
 
+        for each_sg, duplicate_sg_name in SECURITYGROUP_REPLACEMENTS.items():
```
e4r7hbug: I think we need a test case to demonstrate what is supposed to happen. Are we:
* Replacing a Security Group with another Security Group of a different name?
* Removing `key` and adding `value`?
* Removing `value` and adding `key`?

Possible outcomes to clarify:
```python
security_groups = ['a', 'b', 'c', 'd']

replacements = {'a': 'd'}
for key_sg, value_sg in replacements.items():
    …

# Which assertions are correct?
assert security_groups == ['b', 'c']
assert security_groups == ['b', 'c', 'd']
assert security_groups == ['a', 'b', 'c']
assert security_groups == ['a', 'b', 'c', 'd']
```

src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -189,6 +189,11 @@ def ec2_pipeline_setup(appname='', project='', settings=None, env='', region='',
     instance_security_groups.append(appname)
     instance_security_groups.extend(settings['security_group']['instance_extras'])
 
+    for each_sg, duplicate_sg_name in SECURITYGROUP_REPLACEMENTS.items():
```
e4r7hbug: Copying code means this should be a function.

src/foremast/pipeline/construct_pipeline_block_lambda.py
```diff
@@ -68,6 +68,11 @@ def construct_pipeline_block_lambda(env='',
     instance_security_groups.append(gen_app_name)
     instance_security_groups.extend(settings['security_group']['instance_extras'])
 
+    for each_sg, duplicate_sg_name in SECURITYGROUP_REPLACEMENTS.items():
```
e4r7hbug: Third copy means we really need a function.

jvasallo: Done. Moved to utils.

src/foremast/pipeline/construct_pipeline_block.py
jvasallo: Done. Moved to utils.

src/foremast/elb/create_elb.py
jvasallo: Done. Moved to utils.

Writing a test case I suppose around this.

PR #249, opened 2017-12-14
Rolling pipeline type

ndcampbell:
This just makes it so that "rolling" is a valid pipeline type. It needs to do all the same things as ec2 pipelines, except load a different set of templates.


sbasgall:
Seems like a simple enough change. Not much to comment on here.

PR #250, opened 2017-12-14
Return ARN directly if topic name appears to be an ARN

sbasgall:


src/foremast/utils/get_sns_topic_arn.py
```diff
@@ -20,6 +20,8 @@ def get_sns_topic_arn(topic_name, account, region):
         str: ARN for requested topic name
 
     """
+    if topic_name.count(':') == 5:
```
jvasallo: Can you maybe add an additional statement just to do some basic sanitization such as:
`if topic_name.count(':') == 5 and topic_name.contains('arn:aws:sns'):`

sbasgall: we could go all out regex to be absolutely sure

src/foremast/utils/get_sns_topic_arn.py
```diff
@@ -20,7 +20,7 @@ def get_sns_topic_arn(topic_name, account, region):
         str: ARN for requested topic name
 
     """
-    if topic_name.count(':') == 5:
+    if topic_name.count(':') == 5 and topic_name.startswith('arn:aws:sns:')
```
jvasallo: Can you fix the syntax error though ;) `:`

PR #252, opened 2017-12-18
Disable scale down

ndcampbell:
This adds the ability to disable the scale-down policy. Default behavior remains the same.

### DIFF
--- a/_docs/configuration_files/application_json.rst
+++ b/_docs/configuration_files/application_json.rst
@@ -237,6 +237,11 @@ scaling policy will be attached
 
    | *Type*: number
 
+``scale_down`` : Attach a default scale-down policy
+
+   | *Type*: boolean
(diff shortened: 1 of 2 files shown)


sijis:
Is this allowing someone to toggle the scaling functionality?
Do we need to update the default templates too with this new parameter?

ndcampbell:
@sijis currently if someone adds a scaling policy, we add a default scale down policy. This allows you to disable the scale down policy. It is needed for apps like Kafka where they may want to scale up, but never automatically back down or else you'd lose data.

The default template has `asg.scaling_policy = {}` so nothing needs updated there. If we changed that default, we'd have to changed more of the code as we rely on the truthiness of `{}` in a few places.

sijis:
I'm good with this. 👍

PR #253, opened 2017-12-20
Revert "Added support for removing duplicate security group references based on names"

jvasallo:
Reverts gogoair/foremast#248


sijis:
This didn't do what we expected. :/

PR #254, opened 2018-01-08
Add support for Glue and Athena to IAM Services block

jvasallo:
Athena and Glue are a bit different as there is no resource based policies. Typically, you restrict the data behind the scenes with S3 permissions.

https://docs.aws.amazon.com/IAM/latest/UserGuide/list_athena.html
https://docs.aws.amazon.com/glue/latest/dg/api-permissions-reference.html


sijis:
I'm not seeing any issue in this.

PR #255, opened 2018-01-09
Added new tlscert_naming template format to allow support for ACM

jvasallo:
ACM creates certs per region, this change enables users to reference certs by friendly name instead of long ARN. In doing this, we need to have backwards compatibility for both the v1 format and v2. I attempt the lookup in the v1 format first, if not we fall back to v2. 

## Questions
1. Do we want to have an execution flag to default to v2 for internal teams to prevent conflicts? 
2. Working on Unit tests now.

## Sample Template JSON
```{
    "iam": {
        "stage": {
            "wildcard.example.com": "arn::::example"
        },
        "prod": {
            "wildcard.example.com": "arn:::::example"
        }
    },
    "acm": {
        "us-east-1": {
            "stage": {
                "wildcard.appgroup.example.com": "arn:acm:::example"
            },
            "prod": {
                "wildcard.appgroup.example.com": "arn:acm:::example"
            }
        },
        "us-west-2": {
            "stage": {
                "wildcard.appgroup.example.com": "arn:acm:::example"
            },
            "prod": {
                "wildcard.appgroup.example.com": "arn:acm:::example"
            }
        }
    },
    "prod": {
        "wildcard.legacyiam.example.com": "arn:iam:::example",
        "wildcard.legacyacm.example.com": "arn:acm:::example"
    }
}
```

### DIFF
--- a/src/foremast/elb/format_listeners.py
+++ b/src/foremast/elb/format_listeners.py
@@ -178,21 +178,27 @@ def generate_custom_cert_name(env='', region='', account='', certificate=None):
     # TODO: Investigate moving this to a remote API, then fallback to local file if unable to connect
     try:
         rendered_template = get_template(template_file='infrastructure/iam/tlscert_naming.json.j2', **template_kwargs)
+        tlscert_dict = json.loads(rendered_template)
     except ForemastTemplateNotFound:
         LOG.info('Unable to find TLS Cert Template...falling back to default logic...')
         return cert_name
 
     try:
-        cert_name = json.loads(rendered_template)[env][certificate]
+        LOG.info("Attempting to find TLS Cert using TLS Cert Template v1 lookup...")
+        cert_name = tlscert_dict[env][certificate]
+        LOG.info("Found TLS certificate named %s under %s using TLS Cert Template v1", certificate, env)
     except KeyError:
-        LOG.error("Unable to find TLS certificate named %s under %s in v1 TLS Cert Template.", certificate, env)
+        LOG.error("Unable to find TLS certificate named %s under %s using v1 TLS Cert Template.", certificate, env)
 
     if cert_name is None:
-        LOG.info("Attempting v2 TLS Cert Lookup...")
-        if certificate in json.loads(rendered_template)['iam'][env]:
-            cert_name = json.loads(rendered_template)['iam'][env][certificate]
-        elif certificate in json.loads(rendered_template)['acm'][region][env]:
-            cert_name = json.loads(rendered_template)['acm'][region][env][certificate]
+        LOG.info("Attempting to find TLS Cert using TLS Cert Template v2 lookup...")
+        if certificate in tlscert_dict['iam'][env]:
+            cert_name = tlscert_dict['iam'][env][certificate]
+            LOG.info("Found IAM TLS certificate named %s under %s using TLS Cert Template v2", certificate, env)
+        elif certificate in tlscert_dict['acm'][region][env]:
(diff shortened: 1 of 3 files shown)


tests/elb/test_elb.py
```diff
@@ -17,10 +17,54 @@
 import json
 from unittest import mock
 
+import io
```
sijis: I don't believe this is being used.

src/foremast/elb/format_listeners.py
```diff
@@ -177,13 +178,30 @@ def generate_custom_cert_name(env='', account='', certificate=None):
...
     try:
-        cert_name = json.loads(rendered_template)[env][certificate]
+        LOG.info("Attempting to find TLS Cert using TLS Cert Template v1 lookup...")
+        cert_name = tlscert_dict[env][certificate]
+        LOG.info("Found TLS certificate named %s under %s using TLS Cert Template v1", certificate, env)
     except KeyError:
-        LOG.error("Unable to find TLS certificate named %s under %s in TLS Cert Template", certificate, env)
+        LOG.error("Unable to find TLS certificate named %s under %s using v1 TLS Cert Template.", certificate, env)
+
+    if cert_name is None:
```
sijis: You should probably do `if not cert_name`

src/foremast/elb/format_listeners.py
```diff
@@ -159,7 +160,7 @@ def format_cert_name(env='', account='', certificate=None):
     return cert_name
 
 
-def generate_custom_cert_name(env='', account='', certificate=None):
+def generate_custom_cert_name(env='', region='', account='', certificate=None):
     """Generate a custom TLS Cert name based on a template.
 
     Args:
```
sijis: Missing docstring for `account' parameter.

src/foremast/elb/format_listeners.py
```diff
@@ -23,7 +23,7 @@
 LOG = logging.getLogger(__name__)
 
 
-def format_listeners(elb_settings=None, env='dev'):
+def format_listeners(elb_settings=None, env='dev', region='us-east-1'):
```
sijis: Missing docstring regarding `region`.

src/foremast/elb/format_listeners.py
jvasallo: I think you meant region, but fixed.

src/foremast/elb/format_listeners.py
jvasallo: Fixed

tests/elb/test_elb.py
jvasallo: Fixed.

src/foremast/elb/format_listeners.py
sijis: Yeah, `region`. sorry.

jvasallo:
@sijis I have an example in the external repo for templates. I didn't want to necessarily include it cause it will override the default behavior of Foremast to locally craft the name instead of looking up.

https://github.com/gogoair/foremast-template-examples/blob/master/infrastructure/iam/tlscert_naming.json.j2

src/foremast/elb/format_listeners.py
sijis: We discussed offline on this.
@jvasallo felt more comfortable being explicit on checking the value of `cert_name` was `None`. The concern was around if somehow `cert_name` earlier in the function gets a value, it will be match against this condition.
I think its a fair concerns, so we are leaving it as-is.

sijis:
As long as we have an example somewhere we can reference, I'm OK with that.

PR #256, opened 2018-01-10
fix: Cross acount SG within same account

sijis:
We need to ensure that we do not attempt cross account security lookups
when the account is the same as the app. We should only trigger that
lookup when the app and SG are different.


sijis:
It resolves this error
```
raise SpinnakerTaskError(task_state)
foremast.exceptions.SpinnakerTaskError: the specified rule "peer: sg-xx81a9e2, TCP, from port: 443, to port: 443, ALLOW" already exists (Service: AmazonEC2; Status Code: 400; Error Code: InvalidPermission.Duplicate; Request ID: 8c71fbf3-71eb-4efc-984a-7d56e7c16782)
```

jvasallo:
@sijis can you either sync your branch up or fix the linting issue?

jvasallo:
Will rebase and merge as the lint error was fixed in master

PR #257, opened 2018-01-10
Refactor security group ingress rule

sijis:
Try and simplify creating security group code.
This should make it more testable.

Depends on #256.

### DIFF
--- a/src/foremast/securitygroup/create_securitygroup.py
+++ b/src/foremast/securitygroup/create_securitygroup.py
@@ -282,7 +282,6 @@ class SpinnakerSecurityGroup(object):
         self.log.info('Successfully created %s security group', self.app_name)
         return True
 
-
     def create_ingress_rule(self, app, rule):
         """Create a normalized ingress rule.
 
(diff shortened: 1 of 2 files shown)


src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -311,3 +281,48 @@ def create_security_group(self):  # noqa
...
+            app (str): Application name
+            rule (dict or int): Allowed Security Group ports and protocols.
+
+        Returns:
+            dict: Contains app, start_port, end_port, protocol, cross_account_env and cross_account_vpc_id
+
+        """
+        if isinstance(rule, dict):
+            # Advanced
+            start_port = rule.get('start_port', 80)
```
jvasallo: I don't like auto-assuming connectivity in terms of ports. This could accidentally expose port 80 in the event of misconfigured files.

sijis: Good point. Fixed.

PR #260, opened 2018-01-17
refactor: Remove defined types in pipeline_wrapper

sijis:
We already have defined Pipeline types for lambda, s3 and datapipeline.


e4r7hbug:
Great.

PR #262, opened 2018-01-18
refactor: Fallback to default pipeline for unknown types

sijis:
This allows users to create custom pipeline type and not need to make
code updates for those new types to work.


ndcampbell:
I think this looks good to me. We are going to have lots of random pipeline types pop up and this makes it just a template change.

sijis:
We are still restricted based on what's allowed in the configs:
https://github.com/gogoair/foremast/pull/262/files#diff-b72d390d6b9823485abfbaa5031c64e3R104

ndcampbell:
yeah, that is at least still not code changed to add a new type.

PR #265, opened 2018-01-29
fix: Use fully qualified path to file

sijis:
We should show the fully qualified path of the templates that are used.

Old
```
2018-01-29 15:29:18,160 [INFO] foremast.utils.templates:get_template:84 - Rendering template /data/git/github/foremast/src/foremast/utils/../templates/infrastructure/autoscaling_policy.json.j2
```

New
```
2018-01-29 15:22:15,280 [INFO] foremast.utils.templates:get_template:86 - Rendering template /data/git/github/foremast/src/foremast/templates/infrastructure/autoscaling_policy.json.j2
```

### DIFF
--- a/src/foremast/utils/templates.py
+++ b/src/foremast/utils/templates.py
@@ -44,17 +44,17 @@ def get_template_object(template_file=''):
             is not available.
 
     """
-    jinja_lst = []
+    jinja_template_paths_obj = []
 
     if TEMPLATES_PATH:
         external_templates = pathlib.Path(TEMPLATES_PATH).expanduser().resolve()
         assert os.path.isdir(external_templates), 'External template path "{0}" not found'.format(external_templates)
-        jinja_lst.append(external_templates)
+        jinja_template_paths_obj.append(external_templates)
 
-    jinja_lst.append(LOCAL_TEMPLATES)
-    jinja_templates = [str(path) for path in jinja_lst]
+    jinja_template_paths_obj.append(LOCAL_TEMPLATES)
+    jinja_template_paths = [str(path) for path in jinja_template_paths_obj]
 
-    jinjaenv = jinja2.Environment(loader=jinja2.FileSystemLoader(jinja_templates))
+    jinjaenv = jinja2.Environment(loader=jinja2.FileSystemLoader(jinja_template_paths))
 
     try:
         template = jinjaenv.get_template(template_file)
--- a/src/foremast/utils/templates.py
+++ b/src/foremast/utils/templates.py
@@ -25,8 +25,8 @@ from ..exceptions import ForemastTemplateNotFound
 
 LOG = logging.getLogger(__name__)
 
-HERE = os.path.dirname(os.path.realpath(__file__))
-LOCAL_TEMPLATES = pathlib.Path('{0}/../templates/'.format(HERE)).resolve()
+HERE = pathlib.Path(__file__).parent.absolute()
+LOCAL_TEMPLATES = HERE.joinpath('../templates/').resolve()
 
 
 def get_template_object(template_file=''):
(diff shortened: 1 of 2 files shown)


src/foremast/utils/templates.py
```diff
@@ -16,6 +16,7 @@
 """Render Jinja2 template."""
 import logging
 import os
+import pathlib
```
bazimov: Do we have pathlib in requirements.txt?

bazimov: Ah I did not know that. But if we merge that, will it break Python2 users then? How do we keep foremast compatible with Python2?

sijis: We never supported python 2.
https://github.com/gogoair/foremast/blob/master/setup.py#L45

sijis: It if worked, that's great but it was not intentional.

src/foremast/utils/templates.py
```diff
@@ -47,6 +48,7 @@ def get_template_object(template_file=''):
 
     if TEMPLATES_PATH:
         external_templates = os.path.expanduser(TEMPLATES_PATH)
```
e4r7hbug: Delete.

src/foremast/utils/templates.py
```diff
@@ -47,6 +48,7 @@ def get_template_object(template_file=''):
...
+        external_templates = str(pathlib.Path(TEMPLATES_PATH))
```
e4r7hbug: This is missing the previous `.expanduser()` behaviour. It also sounds like you want to start using `.resolve()`.

src/foremast/utils/templates.py
```diff
@@ -25,7 +26,7 @@
 LOG = logging.getLogger(__name__)
 
 HERE = os.path.dirname(os.path.realpath(__file__))
-LOCAL_TEMPLATES = '{0}/../templates/'.format(HERE)
+LOCAL_TEMPLATES = str(pathlib.Path('{0}/../templates/'.format(HERE)).resolve())
```
e4r7hbug: Instead of stringifying as we go, we should wait until the variable is used to stringify.

src/foremast/utils/templates.py
```diff
@@ -25,7 +26,7 @@
 LOG = logging.getLogger(__name__)
 
 HERE = os.path.dirname(os.path.realpath(__file__))
```
e4r7hbug: Convert this to `pathlib` as well, for consistency.

src/foremast/utils/templates.py
sijis: Removed.

src/foremast/utils/templates.py
sijis: Good point. fixed that.

src/foremast/utils/templates.py
sijis: Good idea. I did it.

src/foremast/utils/templates.py
sijis: Refactored.

sijis:
Anything else holding this from being merged?

src/foremast/utils/templates.py
```diff
@@ -46,13 +47,14 @@ def get_template_object(template_file=''):
     jinja_lst = []
 
     if TEMPLATES_PATH:
-        external_templates = os.path.expanduser(TEMPLATES_PATH)
+        external_templates = pathlib.Path(TEMPLATES_PATH).expanduser().resolve()
         assert os.path.isdir(external_templates), 'External template path "{0}" not found'.format(external_templates)
         jinja_lst.append(external_templates)
 
     jinja_lst.append(LOCAL_TEMPLATES)
+    jinja_templates = [str(path) for path in jinja_lst]
```
e4r7hbug: Technically this is a list of Jinja template paths.

sijis: :+1: Updated.

e4r7hbug:
Great.

PR #266, opened 2018-02-07
pass all pipeline data to wrapper creation

ndcampbell:
We do this with stage creation but I need this data during wrapper creation as well so that we can disable a trigger.

This just passes everything in `pipeline.json` to wrapper creation so that extra fields can be added without touching foremast source in the future.

### DIFF
--- a/src/foremast/pipeline/create_pipeline.py
+++ b/src/foremast/pipeline/create_pipeline.py
@@ -127,6 +127,7 @@ class SpinnakerPipeline:
                 'slack': slack,
                 'root_volume_size': root_volume_size,
                 'ami_template_file': ami_template_file,
+                'pipeline': self.settings['pipeline']
             },
             'id': pipeline_id
         }


sijis:
Does this allow all templates access to data in `pipeline.json`?

ndcampbell:
Any stage_* or pipeline_stages_* template already have access to all data in pipeline.json. This would make it so that `pipeline_wrapper.json.j2` has access

src/foremast/pipeline/create_pipeline.py
```diff
@@ -127,6 +127,7 @@ def render_wrapper(self, region='us-east-1'):
                 'slack': slack,
                 'root_volume_size': root_volume_size,
                 'ami_template_file': ami_template_file,
+                'pipeline': self.settings['pipeline']
```
sijis: Would i make sense to add this under its own `pipeline` key, instead as part of `app`?
https://github.com/gogoair/foremast/pull/266/files#diff-f6fa73849891e3ceadc9aed00baa386aR119

ndcampbell: Yeah, that would make sense, except I was just trying to be consistent with the rest of the code. For the pipeline stage blocks we also nest this under app.pipeline

sijis: Would that cause conflict/confusion though?
In 1 case, you have app.pipeline that has environment specific configs (`application-master-$env.json`).
In the other case app.pipeline could be the app leverl configs (`pipeline.json`).

sijis: I'm not saying the approach is wrong, I'm just thinking of any downstream template ramifications in doing this.

sijis: To add, I really like having the contents available in `pipeline.json` to all the templates.

ndcampbell: yeah, this is poor implementation from that past.

`app.pipeline` in both cases contains the same thing, the contents of `pipeline.json`. We render this wrapper template with different variables for some reason.

sijis:
We've discussed this offline and this just adds the same functionality in all the other templates.
All the `app.pipeline` references refer to the contents of `pipeline.json`. This change just adds that missing piece for the pipieline wrapper template.

PR #268, opened 2018-02-13
disabled trigger for other pipelines besides just ec2

ndcampbell:
This was added for ec2 pipelines in https://github.com/gogoair/foremast/pull/266 but forgot to update it in the other pipeline types.


sijis:
Haha. Of course we'd forget that!

PR #271, opened 2018-02-20
Use region specific configs - Lambda and autoscaling 

ndcampbell:
Update lambda and autoscaling to use region-specific configs. If we are cool with these updates, I will do the same for all other services but wanted to start small.

### DIFF
--- a/src/foremast/autoscaling_policy/create_policy.py
+++ b/src/foremast/autoscaling_policy/create_policy.py
@@ -51,7 +51,7 @@ class AutoScalingPolicy:
         self.region = region
         self.app = app
 
-        self.settings = get_properties(properties_file=prop_path, env=self.env)[self.region]
+        self.settings = get_properties(properties_file=prop_path, env=self.env, region=self.region)
 
     def prepare_policy_template(self, scaling_type, period_sec, server_group):
         """Renders scaling policy templates based on configs and variables.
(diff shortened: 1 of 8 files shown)


sijis:
Would something like this make it better?

```
--- a/src/foremast/utils/properties.py
+++ b/src/foremast/utils/properties.py
@@ -20,7 +20,7 @@ import logging
 
 
-def get_properties(properties_file='raw.properties.json', env=None):
+def get_properties(properties_file='raw.properties.json', env=None, region=None):
     """Get contents of _properties_file_ for the _env_.
 
     Args:
@@ -36,5 +36,8 @@ def get_properties(properties_file='raw.properties.json', env=None):
         properties = json.load(file_handle)
 
     env_properties = properties.get(env, properties)
-    LOG.debug('Found properties for %s:\n%s', env, env_properties)
-    return env_properties
+
+    contents = env_properties.get(region, env_properties)
+
+    LOG.debug('Found properties for %s:\n%s\%s', env, env_properties, contents)
+    return contents
```

Then the signature is used as:
```
self.settings = get_properties(properties_file=prop_path, env=self.env, region=self.region)
```

If region is `None`, then you'll get existing behavior, which is the environment level config.

ndcampbell:
@sijis this has been updated with your recommendation. While implementing this I discovered not all the configs were being inserted into the region specific changes and so I fixed that as well.

sijis:
I just ran through some configuration scenarios. I'm not seeing any issues.

PR #272, opened 2018-02-22
Fix failover dns

sijis:
We should only update the global dns record when it matches the primary region.

This addresses a problem where the secondary, region specific, domain was updating the global dns record.


ndcampbell:
Looks good to me. Basic testing seems fine.

PR #276, opened 2018-02-26
Remove reference to invalid stage

sijis:
I included a custom reference to a stage in 2556e13d387b00bc45654db25167f488cf80919e.

The `stage-tagger.json.j2` does not exist and this removes that reference.


ndcampbell:
hah, good catch. LGTM

PR #277, opened 2018-02-28
overrides regions in templates with regions in configs

ndcampbell:
Since we moved to dicts for the `regions` block, the default in our templates gets put in even if excluded in application configs. This makes sure to only use the regions in the application config. If not provided in the application config, it will use the default in the templates.


ndcampbell:
This fixes the following scenario:

If you have `"regions": { "us-east-1": {} }` in the templates as a default, but `"regions": { "us-west-2": {} }` in the application configs, you end up with `"regions": { "us-east-1": {}, "us-west-2": {} }` in the rendered configs due to how `DeepChainMap` works. This makes sure just the default or just the app configs are used

sijis:
Thank you for clarifying the change in the last comment.

I see no issues with this.

PR #280, opened 2018-03-16
Better merge of default and application security group rules.

sijis:
We have the ability add default security group rules via our `config.py` files.

There are scenarios where an application (via `application.json`) needs to add a rule, which correspond to one already defined in `config.py`. In those scenarios, the default definition from `config.py` will win and those definitions will be used instead of what an application has defined.

In does the following:
```
DEFAULT = {'test_app': [{'start_port': 31,'end_port': 31,}]}
APP = {'test_app': [{'start_port': 30,'end_port': 30,}]}

OLD_BEHAVIOR =  {
        'test_app': [
            {
                'start_port': 31,
                'end_port': 31,
            }
        ]
}

NEW_BEHAVIOR =  {
        'test_app': [
            {
                'start_port': 30,
                'end_port': 30,
            },
            {
                'start_port': 31,
                'end_port': 31,
            }
        ]
}

```

### DIFF
--- a/tests/test_securitygroup.py
+++ b/tests/test_securitygroup.py
@@ -37,6 +37,21 @@ SAMPLE_JSON = """{"security_group": {
                     }
                   }}"""
 
+DEFAULT_SG_RULES = {
+    'myapp': [
+        {
+            'start_port': 22,
+            'end_port': 22,
+        },
+    ],
+    'test_app': [
+        {
+            'start_port': 31,
+            'end_port': 31,
+        },
+    ],
+}
+
 
 @mock.patch('foremast.securitygroup.create_securitygroup.boto3')
 @mock.patch('foremast.securitygroup.create_securitygroup.get_security_group_id')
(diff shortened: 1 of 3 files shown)


tests/test_securitygroup.py
```diff
@@ -162,3 +162,45 @@ def test_securitygroup_references(mock_properties, mock_details):
...
+            {
+                'start_port': 22,
+                'end_port': 22,
+            },
+        ],
+        'test_app': [
+            {
+                'start_port': 31,
+                'end_port': 31,
+            }
```
e4r7hbug: More trailing commas.

tests/test_securitygroup.py
```diff
@@ -162,3 +162,45 @@ def test_securitygroup_references(mock_properties, mock_details):
...
+        ]
+    }
+
+    with mock.patch.dict('foremast.securitygroup.create_securitygroup.DEFAULT_SECURITYGROUP_RULES', default_sg):
```
e4r7hbug: Use decorator form.

tests/test_securitygroup.py
```diff
@@ -162,3 +162,45 @@ def test_securitygroup_references(mock_properties, mock_details):
...
+        sg = SpinnakerSecurityGroup()
+        ingress = sg.update_default_rules()
+        assert len(ingress['myapp']) == 1
```
e4r7hbug: Check the values instead of just length. I have a feeling this might get messed up in the future and we'll want to make sure the values are good.

tests/test_securitygroup.py
sijis: fixed.

tests/test_securitygroup.py
sijis: fixed.

tests/test_securitygroup.py
sijis: Good point. Did that.

PR #282, opened 2018-04-06
Flat s3 deployments

ndcampbell:
This enabled the ability to do "flat" s3 deployments, flat meaning no version and no LATEST directory. The contents are uploaded in the exact structure as the artifact. 

This does DELETE everything else in the bucket. It is expected, but potentially dangerous bahavior. Is it worth coming up with some protection? 

This works by simply setting "deploy_strategy" to "flat"  in application-master-env.json.

### DIFF
--- a/src/foremast/s3/s3deploy.py
+++ b/src/foremast/s3/s3deploy.py
@@ -94,10 +94,10 @@ class S3Deployment(object):
     def upload_artifacts(self):
         """Upload artifacts to S3 and copy to correct path depending on strategy."""
         deploy_strategy = self.properties["deploy_strategy"]
+
+        mirror = False
         if deploy_strategy == "mirror":
             mirror = True
-        else:
-            mirror = False
 
         self._upload_artifacts_to_path(mirror=mirror)
         if deploy_strategy == "highlander":
@@ -138,7 +138,8 @@ class S3Deployment(object):
         else:
             dest_uri = self.s3_version_uri
 
-        cmd = 'aws s3 sync {} {} --delete --exact-timestamps --profile {}'.format(self.artifact_path, dest_uri, self.env)
+        cmd = 'aws s3 sync {} {} --delete --exact-timestamps --profile {}'.format(self.artifact_path,
+                                                                                  dest_uri, self.env)
         return cmd
 
     def _upload_artifacts_to_path(self, mirror=False):
--- a/src/foremast/s3/s3deploy.py
+++ b/src/foremast/s3/s3deploy.py
@@ -131,15 +131,15 @@ class S3Deployment(object):
         """
         if flat:
             cmd = 'aws s3 sync {} {} --delete --exact-timestamps --profile {}'.format(self.artifact_path,
-                                                                                    self.s3_flat_uri, self.env)
+                                                                                      self.s3_flat_uri, self.env)
         else:
             cmd = 'aws s3 sync {} {} --delete --exact-timestamps --profile {}'.format(self.artifact_path,
-                                                                                    self.s3_version_uri, self.env)
+                                                                                      self.s3_version_uri, self.env)
         return cmd
 
     def _upload_artifacts_to_path(self, flat=False):
         """Recursively upload directory contents to S3.
-        
+
         Args:
             flat (bool): If true, uses a flat directory structure instead of nesting under a version.
         """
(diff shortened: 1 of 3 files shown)


src/foremast/s3/s3deploy.py
```diff
@@ -79,25 +80,32 @@ def _path_formatter(self, suffix):
...
+        """Upload artifacts to S3 and copy to correct path depending on strategy."""
         deploy_strategy = self.properties["deploy_strategy"]
-        self._upload_artifacts_to_version()
-        if deploy_strategy == "highlander":
-            self._sync_to_uri(self.s3_latest_uri)
-        elif deploy_strategy == "canary":
-            self._sync_to_uri(self.s3_canary_uri)
-        elif deploy_strategy == "alpha":
-            self._sync_to_uri(self.s3_alpha_uri)
+        if deploy_strategy == "flat":
```
sijis: Why is this whole section not like:
```
if deploy_strategy == "x":
    ....
elif deploy_strategy == "y":
   ....
elif deploy_strategy == "z":
   ...
```

src/foremast/s3/s3deploy.py
```diff
@@ -112,36 +120,57 @@ def promote_artifacts(self, promote_stage='latest'):
...
+    def _get_upload_cmd(self, flat=False):
+        """Generate the S3 CLI upload command
+
+        Args:
+            flat (bool): If true, uses a flat directory structure instead of nesting under a version.
+
+        Returns:
+            str: The full CLI command to run.
+        """
+        if flat:
```
sijis: The only difference in the `cmd` variable is that one uses `self.s3_flat_uri` vs `self.s3_version_uri`.
It would be better if it was something like:
```
if flat:
   uri = self.s3_flat
else:
  uri = self.non_s3_flat

cmd = 'aws s3 ....{} {}'.format(self.something, uri)
```

Since there's no difference in the `cmd` except for the path, this makes that more explicitly visible.

src/foremast/s3/s3deploy.py
```diff
@@ -79,25 +80,32 @@ def _path_formatter(self, suffix):
             str: formatted path
 
         """
-        path_format = "{}/{}/{}"
+        if suffix.lower() == "flat":
```
sijis: I'm wondering if this makes it more legible:
```
if flat:
  path = [self.bucket, self.s3path]
else:
  path = [self.bucket, self.s3path, suffix]

path_format = '/'.join(path)
```

sijis:
What sort of 'protection' are you thinking about? Notifying folks that a delete will happen or something?

src/foremast/s3/s3deploy.py
ndcampbell: because of the `  self._upload_artifacts_to_path(flat=False)` part should run before all the other non-flat deployments.

src/foremast/s3/s3deploy.py
ndcampbell: yeah, I like that, will update

src/foremast/s3/s3deploy.py
ndcampbell: yeah, that makes sense. I will update it

ndcampbell:
@sijis No idea on protection. I personal feel that no protection is necessary, we can make it clear in the docs that this is what will happen. Just worth noting this could be a destructive deployment. 

Also, I'm not a huge fan of `flat` either, but I cannot think of anything better. Maybe `mirror` could work too if you like that better

src/foremast/s3/s3deploy.py
sijis: Ahh. So no matter what `self._upload_artifacts_to_path(..)` needs to run. If that's so, then i think this would work:

```
flat = False
if deploy_strategy = 'flat':
   flat = True

# this runs against all strategies
self._upload_artifacts_to_path(flat=flat)

# run specific sync_uri function
if deploy_strategy == "x":
     self._sync_to_uri(self.s3_latest_uri)
elif deploy_strategy == "y":
    self._sync_to_uri(self.s3_canary)
elif deploy_strategy == "z":
   ....

```

ndcampbell: ahh yeah, I like that.

sijis:
I like all the changes going on. I think we can add some simple unit tests around these changes too.

ndcampbell:
@sijis - cool, will add some tests today

sijis:
No issues. Thakns for adding tests and docs.

PR #283, opened 2018-04-09
Updated Athena and Glue permissions

jvasallo:
- Athena now includes glue:Get* as this is now required once Athena's data catalog is upgraded.
- Glue was converted to generic Get* as this might need to be tweaked in
the future

### DIFF
--- a/src/foremast/templates/infrastructure/iam/glue.json.j2
+++ b/src/foremast/templates/infrastructure/iam/glue.json.j2
@@ -1,8 +1,7 @@
         {
             "Sid": "GlueLimitedAccess",
             "Action": [
-                "glue:GetTable*",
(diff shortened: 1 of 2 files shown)


src/foremast/templates/infrastructure/iam/glue.json.j2
```diff
@@ -1,8 +1,7 @@
         {
             "Sid": "GlueLimitedAccess",
             "Action": [
-                "glue:GetTable*",
-                "glue:GetDatabase*"
+                "glue:Get*"
```
sijis: Do you think `Get*` is too broad?

sijis: Fair. Damn AWS.

PR #284, opened 2018-04-09
docs: add all runtime strings

sbasgall:
Update lambda docs to include all valid runtime strings

### DIFF
--- a/_docs/configuration_files/pipeline_json/lambda.rest
+++ b/_docs/configuration_files/pipeline_json/lambda.rest
@@ -25,8 +25,13 @@ The runtime environment for the Lambda function
         - ``"java8"``
         - ``"nodejs"``
         - ``"nodejs4.3"``
+        - ``"nodejs6.10"``
         - ``"python2.7"``
         - ``"python3.6"``
+        - ``"dotnetcore1.0"``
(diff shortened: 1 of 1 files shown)


_docs/configuration_files/pipeline_json/lambda.rest
```diff
@@ -25,8 +25,13 @@ The runtime environment for the Lambda function
         - ``"java8"``
         - ``"nodejs"``
         - ``"nodejs4.3"``
+        - ``"nodejs6.10"``
```
jvasallo: Can we add:
- nodejs8.10

Also maybe have a link: https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html

Other than that, looks good!

sbasgall: nodejs8.10 is not included in the aws lambda api documentation - https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html

have we tested it?

jvasallo: It's documented so it should work. Also the announcement is out too. If you don't want to add it, it is fine :) We can add it later when the documentation is not in flux.

PR #286, opened 2018-04-13
Updates bucket naming to include region if not primary region

ndcampbell:
This PR will use different naming convention for region specific buckets outside the primary region. The reason the primary region does not have a different name is for backwards compatibility.

### DIFF
--- a/src/foremast/runner.py
+++ b/src/foremast/runner.py
@@ -145,25 +145,29 @@ class ForemastRunner(object):
     def deploy_s3app(self):
         """Deploys artifacts contents to S3 bucket"""
         utils.banner("Deploying S3 App")
+        primary_region = self.configs['pipeline']['primary_region']
         s3obj = s3.S3Deployment(
             app=self.app,
             env=self.env,
             region=self.region,
             prop_path=self.json_path,
             artifact_path=self.artifact_path,
-            artifact_version=self.artifact_version)
+            artifact_version=self.artifact_version,
+            primary_region=primary_region)
         s3obj.upload_artifacts()
 
     def promote_s3app(self):
         """promotes S3 deployment to LATEST"""
         utils.banner("Promoting S3 App")
+        primary_region = self.configs['pipeline']['primary_region']
         s3obj = s3.S3Deployment(
             app=self.app,
             env=self.env,
             region=self.region,
             prop_path=self.json_path,
             artifact_path=self.artifact_path,
-            artifact_version=self.artifact_version)
+            artifact_version=self.artifact_version,
+            primary_region=primary_region)
         s3obj.promote_artifacts(promote_stage=self.promote_stage)
 
     def create_secgroups(self):
--- a/src/foremast/runner.py
+++ b/src/foremast/runner.py
@@ -134,7 +134,8 @@ class ForemastRunner(object):
     def create_s3app(self):
         """Create S3 infra for s3 applications"""
         utils.banner("Creating S3 App Infrastructure")
-        s3obj = s3.S3Apps(app=self.app, env=self.env, region=self.region, prop_path=self.json_path)
+        primary_region = self.configs['pipeline']['primary_region']
+        s3obj = s3.S3Apps(app=self.app, env=self.env, region=self.region, prop_path=self.json_path, primary_region=primary_region)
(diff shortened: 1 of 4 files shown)


src/foremast/runner.py
```diff
@@ -134,7 +134,12 @@ def create_archaius(self):
     def create_s3app(self):
         """Create S3 infra for s3 applications"""
         utils.banner("Creating S3 App Infrastructure")
-        s3obj = s3.S3Apps(app=self.app, env=self.env, region=self.region, prop_path=self.json_path)
+        primary_region = self.configs['pipeline']['primary_region']
```
jvasallo: can we call this default_region? We can then use this setting as a default similar how AWS CLI works :)

src/foremast/s3/s3apps.py
```diff
@@ -30,33 +30,37 @@
...
-    def __init__(self, app, env, region, prop_path):
+    def __init__(self, app, env, region, prop_path, primary_region='us-east-1'):
         """S3 application object. Setups Bucket and policies for S3 applications.
 
         Args:
             app (str): Application name
             env (str): Environment/Account
             region (str): AWS Region
             prop_path (str): Path of environment property file
+            primary_region (str): The primary region for the application.
```
jvasallo: Same as above, default_region makes it a bit more like AWS CLI (in that you can specify a default region if none specified)

tests/s3/test_s3deploy.py
```diff
@@ -21,7 +21,6 @@ def s3deployment(mock_get_details, mock_get_props):
                                 artifact_version="1")
     return deployobj
 
-
```
jvasallo: Oh the horror. Thanks for cleaning this up :)

src/foremast/s3/s3apps.py
```diff
@@ -67,7 +71,11 @@ def create_bucket(self):
                 LOG.error("Shared bucket %s does not exist", self.bucket)
                 raise S3SharedBucketNotFound
         else:
-            _response = self.s3client.create_bucket(ACL=self.s3props['bucket_acl'], Bucket=self.bucket)
+            if self.region == 'us-east-1':
```
jvasallo: Could this leverage the variable default_region? Would be helpful for other users who maybe have us-west-2 as their default for example.

src/foremast/s3/s3apps.py
```diff
@@ -67,7 +71,11 @@ def create_bucket(self):
...
+                _response = self.s3client.create_bucket(ACL=self.s3props['bucket_acl'], Bucket=self.bucket)
+            else:
+                _response = self.s3client.create_bucket(ACL=self.s3props['bucket_acl'], Bucket=self.bucket,
+                                                        CreateBucketConfiguration={'LocationConstraint': self.region})
```
jvasallo: Is there an issue if we pass LocationConstraint for any region?

src/foremast/runner.py
ndcampbell: we just call it "primary_region" in the configs, so I stuck with that terminology here. It is a value you can override in pipeline.json and originally was used for the multi-region DNS stuff.

jvasallo: Got it. What we called it elsewhere so lets remain consistent. 👍

src/foremast/s3/s3apps.py
jvasallo: Disregard!

src/foremast/s3/s3apps.py
ndcampbell: oh duh, just missed that. Didnt want us-east-1 hardcoded anywhere

src/foremast/s3/s3apps.py
ndcampbell: lol yeah, "us-east-1" does not work in there. All other regions are valid, but it needs to exclude for the "standard" region.

src/foremast/s3/s3apps.py
ndcampbell: Oh, actually us-east-1 is hardcoded on purpose. It is the only value where it needs to be true, even if the default region is something else, you need to include the `CreateBucketConfiguration` part of the command.

PR #287, opened 2018-04-16
fixed issue with non-east buckets already existing and giving errors

ndcampbell:
checks that non-east buckets exist before trying to create. 

https://stackoverflow.com/questions/10346378/boto-s3-error-bucketalreadyownedbyyou description of why this logic is necessary.


jvasallo:
Looks like a lint issue but other than that lgtm!

PR #288, opened 2018-04-18
refactor: Support overriding EC2 Pipeline Types

e4r7hbug:
Need a way to inject custom Pipeline Types that are deployed with EC2.

### DIFF
--- a/_docs/configuration_files/foremast_config.rst
+++ b/_docs/configuration_files/foremast_config.rst
@@ -141,6 +141,12 @@ security group name, it will self-reference to its own application name.
     | *Required*: No
     | *Example*: ``{ "bastion" : [ { "start_port": "22", "end_port": "22", "protocol": "tcp" } ] }``
 
+``ec2_pipeline_types``
+**********************
+
+.. autodata:: foremast.consts.EC2_PIPELINE_TYPES
+   :noindex:
+
 ``gate_client_cert``
 ********************
 
--- a/src/foremast/consts.py
+++ b/src/foremast/consts.py
@@ -201,6 +201,19 @@ AMI_JSON_URL = validate_key_values(CONFIG, 'base', 'ami_json_url')
 DEFAULT_SECURITYGROUP_RULES = _generate_security_groups('default_securitygroup_rules')
 DEFAULT_EC2_SECURITYGROUPS = _generate_security_groups('default_ec2_securitygroups')
 DEFAULT_ELB_SECURITYGROUPS = _generate_security_groups('default_elb_securitygroups')
+
+EC2_PIPELINE_TYPES = tuple(validate_key_values(CONFIG, 'base', 'ec2_pipeline_types', default='ec2,rolling').split(','))
+"""Comma separated list of Pipeline Types to treat as EC2 deployments.
+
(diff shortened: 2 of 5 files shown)


sijis:
Does this not fit the need?
http://foremast.readthedocs.io/en/latest/configuration_files/foremast_config.html?highlight=types#id11

Is your expectation to have a subset of that list or tie those pipelines to a specific deployment 'type' (ec2, lambda, etc)?

e4r7hbug:
My EC2 Pipeline Types will be a subset of `types`. Just like the `rolling` Type, my custom Pipeline Type will be tied to the EC2 deployment style, but will be using different templates.

If the modified conditions in my Commit used `types`, then something like `lambda` or `container` would trigger EC2 related data lookups, which wouldn't make sense.

e4r7hbug:
Looks like I still need this. The hardcoded `('ec2', 'rolling')` Types are limiting. Unless we're ready to push Foremast 4 and switch to only `config.py`, supporting Type aliases in the old INI format will be very cumbersome.

sijis:
In thinking about this, i think its OK for now. Its very concise and explicit in its usage.
I don't think we are ready for a foremast 4.x release as we'd have to start sprinkling a lot of deprecation notices. Once we do that, we can refactor and move some of these things into better names/configs/etc. At the same time, we can remove the INI config.

_docs/configuration_files/foremast_config.rst
```diff
@@ -141,6 +141,12 @@ security group name, it will self-reference to its own application name.
     | *Required*: No
     | *Example*: ``{ "bastion" : [ { "start_port": "22", "end_port": "22", "protocol": "tcp" } ] }``
 
+``ec2_pipeline_types``
+**********************
+
+.. autodata:: foremast.consts.EC2_PIPELINE_TYPES
```
sijis: So why not simply add the content here as the rest of the docs?

src/foremast/consts.py
```diff
@@ -207,6 +207,19 @@ def _generate_security_groups(config_key):
 DEFAULT_SECURITYGROUP_RULES = _generate_security_groups('default_securitygroup_rules')
 DEFAULT_EC2_SECURITYGROUPS = _generate_security_groups('default_ec2_securitygroups')
 DEFAULT_ELB_SECURITYGROUPS = _generate_security_groups('default_elb_securitygroups')
+
+EC2_PIPELINE_TYPES = tuple(validate_key_values(CONFIG, 'base', 'ec2_pipeline_types', default='ec2,rolling').split(','))
+"""Comma separated list of Pipeline Types to treat as EC2 deployments.
```
sijis: See note about `autodata` usage.

_docs/configuration_files/foremast_config.rst
e4r7hbug: I've been trying to get us to move the documentation into the code for a while. This way, if the code needs to change, the documentation is right there to update. In particular with the configuration documentation, the defaults are more likely to match the code reality. The secondary benefit is the `automodule` generated documentation also has the information when looking at the source documentation.

e4r7hbug:
Added simplistic test case.

sijis:
All good.

PR #289, opened 2018-04-24
Support for S3 Lifecycle Rules and Rosco Bake Tweaks

jvasallo:


_docs/configuration_files/s3.rest
```diff
@@ -152,6 +152,78 @@ For more details refer to: http://boto3.readthedocs.io/en/latest/reference/servi
...
+                    },
+                ],
+                'NoncurrentVersionExpiration': {
+                    'NoncurrentDays': 123
+                },
+                'AbortIncompleteMultipartUpload': {
+                    'DaysAfterInitiation': 123
+                }
+            }
+          ]
```
bazimov: I see above is aws docs and simple to understand but do we want to provide an example policy as well? One simple example?

sijis: I think this example is missing the top key maybe?

```
{
  "lifecycle_rules": [
      { "Expiration": ... }
    ],
    [ 
      { "Expiration": ... }
    ]
}
```

src/foremast/s3/s3apps.py
```diff
@@ -189,6 +190,20 @@ def _put_bucket_encryption(self):
         LOG.debug('Response setting up S3 encryption: %s', _response)
         LOG.info('S3 encryption configuration updated')
 
+    def _put_bucket_lifecycle(self):
+        if self.s3props['lifecycle']['enabled']:
+            lifecycle_config = {'Rules': [{}]}
```
sijis: Why is this line needed? Couldn't you do some sort of `dict.update()`?

src/foremast/s3/s3apps.py
```diff
@@ -189,6 +190,20 @@ def _put_bucket_encryption(self):
...
+            lifecycle_config = {
+                'Rules': self.s3props['lifecycle']['lifecycle_rules']
+            }
+            LOG.debug(lifecycle_config)
```
sijis: You should prefix the debug with what it is...
`LOG.debug('Lifecycle Config: ....)`

src/foremast/s3/s3apps.py
```diff
@@ -189,6 +190,20 @@ def _put_bucket_encryption(self):
...
+            _response = self.s3client.put_bucket_lifecycle_configuration(Bucket=self.bucket,
+                                                                         LifecycleConfiguration=lifecycle_config)
+        else:
+            _response = self.s3client.delete_bucket_lifecycle(Bucket=self.bucket)
+        LOG.debug('Response setting up S3 lifecycle: %s', _response)
+        LOG.info('S3 lifecycle configuration updated')
```
sijis: Is this line always true for apps that have lifecycle disabled?

_docs/configuration_files/pipeline_json/image.rest
```diff
@@ -3,6 +3,15 @@
 
 Holds settings for the baked image
 
+``bake_instance_type``
+**********************
+
+Defines the instance type for Rosco (bake step) to use. This could help with issues
+of large and complex bakes.
```
sijis: Do we want to add the link to ec2info like we do for `instance_type`?

src/foremast/s3/s3apps.py
```diff
@@ -189,6 +190,20 @@ def _put_bucket_encryption(self):
         LOG.debug('Response setting up S3 encryption: %s', _response)
         LOG.info('S3 encryption configuration updated')
 
+    def _put_bucket_lifecycle(self):
```
sijis: Missing docstring

src/foremast/s3/s3apps.py
```diff
@@ -189,6 +190,20 @@ def _put_bucket_encryption(self):
...
+                'Rules': self.s3props['lifecycle']['lifecycle_rules']
+            }
+            LOG.debug(lifecycle_config)
+            _response = self.s3client.put_bucket_lifecycle_configuration(Bucket=self.bucket,
+                                                                         LifecycleConfiguration=lifecycle_config)
+        else:
+            _response = self.s3client.delete_bucket_lifecycle(Bucket=self.bucket)
+        LOG.debug('Response setting up S3 lifecycle: %s', _response)
+        LOG.info('S3 lifecycle configuration updated')
+
```
sijis: You should probably return `_response`.

_docs/configuration_files/s3.rest
ntangsuratupgrade: @sijis, the top key is missing on purpose, which is misleading. It's handled in the Function, https://github.com/gogoair/foremast/pull/289/files?diff=unified#diff-29843ed9507806c7194f2bf847faccd3R197.

And I agree with @bazimov, I need a simple, sane copy and paste example in JSON (the current example is in broken Python).

_docs/configuration_files/s3.rest
```diff
@@ -152,6 +152,78 @@ For more details refer to: http://boto3.readthedocs.io/en/latest/reference/servi
...
+    | *Type*: boolean
+    | *Default*: ``false``
+
+``lifecycle_rules`` :
+
+A list of S3 lifecycle rules, if a lifecycle exists it replaces it.
+For more details refer to: http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_bucket_lifecycle_configuration
+
+    | *Type*: object
+    | *Default*: ``None``
```
ntangsuratupgrade: Use the real default, currently `[{}]`.

_docs/configuration_files/s3.rest
```diff
@@ -152,6 +152,78 @@ For more details refer to: http://boto3.readthedocs.io/en/latest/reference/servi
...
+
+    | *Type*: boolean
+    | *Default*: ``false``
+
+``lifecycle_rules`` :
+
+A list of S3 lifecycle rules, if a lifecycle exists it replaces it.
+For more details refer to: http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_bucket_lifecycle_configuration
+
+    | *Type*: object
```
ntangsuratupgrade: The Type required is `array`.

_docs/configuration_files/pipeline_json/image.rest
jvasallo: Linking to official AWS docs.

src/foremast/s3/s3apps.py
jvasallo: Fixed.

_docs/configuration_files/s3.rest
jvasallo: I either opt for leaving it as is, or removing it completely. There are far too many fields to provide an accurate example. Not including all the potential options of a single rule can be further misleading to users. Ultimately, this should be sources from the docs so I am leaning more towards removing it.

_docs/configuration_files/pipeline_json/image.rest
```diff
@@ -3,6 +3,16 @@
...
+Refer to: https://aws.amazon.com/ec2/instance-types/ 
+
+    | *Type*: string
+    | *Default*: `"t2.small"`
```
e4r7hbug: Not enough backticks.

src/foremast/s3/s3apps.py
```diff
@@ -189,6 +189,22 @@ def _put_bucket_encryption(self):
...
+                'Rules': self.s3props['lifecycle']['lifecycle_rules']
+            }
+            LOG.debug('Lifecycle Config: %s', lifecycle_config)
+            _response = self.s3client.put_bucket_lifecycle_configuration(Bucket=self.bucket,
+                                                                         LifecycleConfiguration=lifecycle_config)
+            status = 'applied'
+        else:
+            _response = self.s3client.delete_bucket_lifecycle(Bucket=self.bucket)
+        LOG.debug('Response setting up S3 lifecycle: %s', _response)
+        LOG.info('S3 lifecycle configuration %s', status)
```
e4r7hbug: I still want what @sijis suggested with returning anything possibly useful. Looking at `boto3` documentation, `_response` will always be `None` at the moment. In lieu of that, please return the `status` for now.

_docs/configuration_files/pipeline_json/image.rest
jvasallo: Fixed in Master

src/foremast/s3/s3apps.py
jvasallo: The goal was to stylistically keep it similar to the existing source. If we want to retool returns and responses, we can. But that should be handled outside of this PR. In addition, I will be returning nothing with no use cases currently. If we had a use case sure, but today there was no use case.

PR #290, opened 2018-04-25
pin botocore to fix pypi caching issues

ndcampbell:
We keep getting bit by botocore updating and the pypi caches not updating. Lets pin the version for now and re-evaluate in the future


ndcampbell:
Closing, decided to go a different way with this instead of maintaining pinned versions in Foremast

PR #292, opened 2018-05-01
Converted Archaius to feature toggle to remove assumption of internal logic

jvasallo:
Not everyone uses archaius in the community, let alone uses an application.properties. This at least converts this logic to a toggle until this can be investigated further in a future release.

### DIFF
--- a/_docs/configuration_files/application_json.rst
+++ b/_docs/configuration_files/application_json.rst
@@ -36,6 +36,17 @@ Describes the application.
     | *Type*: string
     | *Default*: ``null``
 
+.. _archaius_enabled:
+
+``archaius_enabled``
+******************
+
+Setting this value to ``true`` will autocreate archiaus pathing in
+a specified archaius S3 bucket.
+
+    | *Type*: boolean
+    | *Default*: ``false``
+
 .. _eureka_enabled:
 
 ``eureka_enabled``
--- a/src/foremast/runner.py
+++ b/src/foremast/runner.py
@@ -254,12 +254,15 @@ def prepare_infrastructure():
     runner.write_configs()
     runner.create_app()
 
+    archaius = runner.configs[runner.env]['app']['archaius_enabled']
     eureka = runner.configs[runner.env]['app']['eureka_enabled']
     deploy_type = runner.configs['pipeline']['type']
 
     if deploy_type not in ['s3', 'datapipeline']:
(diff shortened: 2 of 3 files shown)


ndcampbell:
This could potentially be a breaking change. It is changing default behavior and would require us to rebuild all of our pipelines to get a True value. Maybe the default should be True to at least keep existing functionality?

jvasallo:
Since prepare-infra is a external step, when this runs it auto generates from the configs.

I agree with the comment about this being a breaking change. I argue that the OSS community most likely does not implement Archaius in the way we do, and if they do, they did it in an opinionated way because our internal logic leak. This can be seen from the issue referenced and also the lack of documentation as to why a random file must be created. This at least unifies both internal and external expectations allowing external users to carry on with our assumptions. This is also fixed in an internal template change :D

sijis:
What's the status of this?

jvasallo:
Oh look at that! I need this 👍Thanks past Joel!

@sijis I think we can merge this pending review from someone at Gogo to make sure my comment about:

> This is also fixed in an internal template change to default to True.

I ran this branch today against my Spinnaker cluster and it properly skipped the Archaius steps as expected when ran by default.

@darinpantley and @sbasgall When you folks get a chance, can you review that in your internal templates (some-gogo-foremast-templates/configs/configs.json.j2) the line:
```archaius_enabled: true,```
is present?

sijis:
@jvasallo Just change the default to `True` and nothing needs to be done for existing users.
We should try and keep existing behavior as much as possible.

jvasallo:
@sijis Understood but expecting all foremast users to have an Archaius S3 Bucket is a flawed concept I feel; especially since not everyone uses it like we most likely did.

We will end up with poor usability experiences such as:
https://github.com/foremast/foremast/issues/187

jvasallo:
I was debating the same thing. Either this causes a bump to v4.0.0 or at the least v3.52.0 (from v3.51.6).

Seeing as there is a configurable toggle with a backwards compatible support, not sure how breaking it really is. It can go either way. We could always do what Sijis said and default to True; which I disagree with personally. This should be treated the same way we treat eureka support in my opinion; if a consumer has it, they should enable it, but majority probably don't

sijis:
Make it a `v3.52.0` and move on.

PR #299, opened 2018-06-06
fix: Use dict configs

e4r7hbug:
Need to start using the configurations in the `dict` form. The way `configparser.ConfigParser` works, it squashes all configuration values into `str` and negates some of the benefits of using Dynamic Configurations with `config.py`.

In particular, I'm concerned with the Application Format configurations in `config.py`. Because I want to use some bizarre string formatting, I need to override the `str.format()` behaviour using a Subclass.

```python
# config.py

class AppName(str):
    """Custom Application name manipulation."""

    def format(self, *args, **kwargs):
        """Override default formatting."""
        repo = kwargs['repo']
        first, *rest = repo.split('-')
        camel = [word.capitalize() for word in rest]

        app_name = ''.join([first, *camel])

        return super().format(*args, app_name=app_name, **kwargs)

CONFIG = {
    'formats': {
        'app': AppName('{app_name}'),
    },
}
```

But because `configparser.ConfigParser` squashes everything (essentially running everything through `str()`), I end up with:

```python
CONFIG = {
    'formats': {
        'app': '{app_name}',
    },
}
```

In an effort to deprecate INI style configurations, we need to remove `configparser.ConfigParser` specific interactions. This also simplifies some parts and avoids side effects in Functions.

### DIFF
--- a/src/foremast/consts.py
+++ b/src/foremast/consts.py
@@ -87,8 +87,10 @@ def extract_formats(config_handle):
     return formats
 
 
-def load_dynamic_config(configurations, config_dir=getcwd()):
+def load_dynamic_config(config_dir=getcwd()):
     """Load and parse dynamic config"""
+    dynamic_configurations = {}
+
     # Create full path of config
     config_file = '{path}/config.py'.format(path=config_dir)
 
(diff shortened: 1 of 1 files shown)


sijis:
Do you think there is some deficiencies in `gogo-utils` in regards to this?

Lastly, could you re-push so that it triggers CI?

e4r7hbug:
I don't think `gogo-utils` is limited. It's more of the way we use it in Foremast that makes it hard to patch what I want.
* If we had only one `gogoutils.Generator()`, then I could patch the `app_name` Method
* Monkey patching at the global Class level is probably not what I want to do
* As demonstrated, `gogo-utils` would behave as expected if it weren't for the behaviour of `configparser.ConfigParser`

e4r7hbug:
If we updated `gogo-utils` to support Functions, I could pass in a formatting Function instead, but we would still need this patch to allow Functions to be passed from `config.py > foremast > gogoutils.Generator > gogoutils.Formats`.

sijis:
Ugh. I've always liked the simplicity of `gogo-utils`.
Fixing the `configparser.ConfigParser` behavior seems like the best approach.

Issue #300, opened 2018-06-07
Travis Tests Broke after Moving to Foremast Org?


jvasallo:
Looks like foremast tests are no longer running after we moved the org over. We should ensure that both Travis and ReadTheDocs continue to work.

sijis:
I think it fixed Travis.
My test PR seem to be going through the paces now
https://github.com/foremast/foremast/pull/301

e4r7hbug:
Checking Read the Docs now.

jvasallo:
Cool stuff. Thanks folks. Saw this last night. :)

e4r7hbug:
There's this random Profile that's a Maintainer on Read the Docs, https://readthedocs.org/profiles/gogoair/.

jvasallo:
@e4r7hbug I have that login, Does it make sense to transition this to a generic account we own as an org?

e4r7hbug:
We could, but I don't think we need it right now, considering all Maintainers have admin privileges.

e4r7hbug:
We should also disable any Jenkins Jobs and start using Travis for releasing to PyPI.

jvasallo:
@e4r7hbug 👍 to that.

sijis:
@e4r7hbug if you configure that, we can disable that Jenkins job here that's currently doing it.

jvasallo:
@e4r7hbug @sijis This is resolved I believe; right?

sijis:
Should be.

jvasallo:
Thanks! :)

PR #302, opened 2018-06-08
feat: Deploy with Travis

e4r7hbug:
One `deploy` block will deploy Commits to the `master` Branch. The other block will deploy Tags. I can't figure out a way to combine these into one block due to the way Travis `deploy` works where specifying `tags` ignores `branch` and specifying `branch` ignores `tags`.

There is a small race condition if Tags are pushed faster than Travis can start building the Branch Commit. This results in both Branch and Tag builds trying to deploy the same Tagged version and one will show as false failure.

Example false failure:
* `master` Commit build created: https://travis-ci.org/e4r7hbug/foremast/builds/389587423
    ```
    Uploading distributions to https://test.pypi.org/legacy/
    Uploading foremast-3.46.3-py2.py3-none-any.whl
    100% 206k/206k [00:01<00:00, 109kB/s]
    ```
* Tag build: https://travis-ci.org/e4r7hbug/foremast/builds/389587428
    ```
    Uploading distributions to https://test.pypi.org/legacy/
    Uploading foremast-3.46.3-py2.py3-none-any.whl
    100% 206k/206k [00:00<00:00, 331kB/s]
    HTTPError: 400 Client Error: File already exists. See https://test.pypi.org/help/#file-name-reuse for url: https://test.pypi.org/legacy/
    ```

See also: #300

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,11 @@ notifications:
     on_success: never
     on_failure: always
     on_start: never
+deploy:
+  provider: pypi
+  user: foremast-deployer
+  password:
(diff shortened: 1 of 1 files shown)


sijis:
@e4r7hbug Do you want to try this on the next release?

PR #303, opened 2018-06-13
docs: Fixed wrong key in lambda_environment example for application json

MilosRasic:
Just a quick small docs fix. The key used in the example is "environment" but should be "lambda_environment".


sijis:
@MilosRasic Great catch. Thanks!

PR #307, opened 2018-07-18
Added support for Fiat Permissions and Traffic Guards

jvasallo:


src/foremast/templates/infrastructure/app_data.json.j2
```diff
@@ -12,25 +12,40 @@
             "repoProjectKey": "{{ appinfo.project }}",
             "repoSlug": "{{ appinfo.repo }}",
             "platformHealthOnlyShowOverride": true,
+            "permissions": {
+                "READ": [{% for role_name in pipeline_config.permissions.read_roles %}
```
e4r7hbug: Does `"READ": {{ pipeline_config.permissions.read_roles | tojson }}` work?

sijis: If that does work, it would definitely simplify the template with no loops and if/else condition for last comma.

jvasallo: I like it, let's try it.

PR #308, opened 2018-07-18
Support to run pipelines using a default `Run As User`

jvasallo:
- Needed for Fiat enabled applications to specify service account

### DIFF
--- a/_docs/configuration_files/foremast_config.rst
+++ b/_docs/configuration_files/foremast_config.rst
@@ -107,6 +107,15 @@ relative to where you where you are running the Foremast commands. See
 
     | *Required*: No
 
+``default_run_as_user``
+***********************
+
+Default user to run pipelines as. This is needed for leveraging service accounts in Fiat.
+
+    | *Type*: str
+    | *Default*: ``''``
+    | *Required*: No
+
 ``default_ec2_securitygroups``
 ******************************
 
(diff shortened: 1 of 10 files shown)


_docs/configuration_files/foremast_config.rst
```diff
@@ -107,6 +107,15 @@ relative to where you where you are running the Foremast commands. See
 
     | *Required*: No
 
+``default_run_as_user``
+***********************
+
+Default user to run pipelines as. This is needed for leveraging service accounts in Fiat.
+
+    | *Type*: str
+    | *Default*: ``''``
```
e4r7hbug: Default should be `null`.

src/foremast/templates/pipeline/trigger-jenkins.json.j2
```diff
@@ -4,5 +4,6 @@
       "type": "jenkins",
       "master": "JenkinsCI",
       "job": "{{ data.app.triggerjob }}",
+      "runAsUser": "{{ data.app.run_as_user }}",
```
sijis: For users that are not using fiat, the current default of `''` will likely cause issues. I see @e4r7hbug thought on making this a `null`/`None` default.
Alternatively, its creating a jinja conditional if `data.app.run_as_user` is populated or not.

savilesupgrade:
:+1: for latest changes.

PR #311, opened 2018-07-22
Send `gogoutils.Generator` to templates

sijis:
This allows the object to be accessible to some templates via the `generated` object in jinja.

For example, in `templates/configs/configs.json.j2`, something like this can be added:
```
"lambda_role": {{ generated.app_name() }},
```
This does depend on version 1.9.0 and above of `gogo-utils`.

### DIFF
--- a/src/foremast/configs/outputs.py
+++ b/src/foremast/configs/outputs.py
@@ -91,7 +91,7 @@ def write_variables(app_configs=None, out_file='', git_short=''):
                     env=env,
                     app=generated.app_name(),
                     profile=instance_profile,
-                    generated=generated))
+                    formats=generated))
             json_configs[env] = dict(DeepChainMap(configs, rendered_configs))
             region_list = configs.get('regions', rendered_configs['regions'])
             json_configs[env]['regions'] = region_list  # removes regions defined in templates but not configs.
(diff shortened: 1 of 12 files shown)


src/foremast/configs/outputs.py
```diff
@@ -86,15 +86,20 @@ def write_variables(app_configs=None, out_file='', git_short=''):
         if env != 'pipeline':
             instance_profile = generated.iam()['profile']
             rendered_configs = json.loads(
-                get_template('configs/configs.json.j2', env=env, app=generated.app_name(), profile=instance_profile))
+                get_template(
+                    'configs/configs.json.j2',
+                    env=env,
+                    app=generated.app_name(),
+                    profile=instance_profile,
+                    generated=generated))
```
jvasallo: Can we make this more clear to what it is? `generated_data` or `templated_data` or something along those lines? I think it will make it easier long term. Feel free to discuss.

src/foremast/securitygroup/create_securitygroup.py
```diff
@@ -71,8 +71,8 @@ def __init__(self, app=None, env=None, region=None, prop_path=None):
         self.region = region
 
         self.properties = get_properties(properties_file=prop_path, env=self.env, region=self.region)
-        generated = get_details(app=self.app_name)
-        self.group = generated.data['project']
+        self.generated = get_details(app=self.app_name)
```
jvasallo: Not sure if there will ever be a case where `self.generated` != `generated`. Can't think of anything of the top of my head, so might be a non-issue, but @e4r7hbug thoughts?

src/foremast/configs/outputs.py
savilesupgrade: I think that is a valid point. I just wanted to have something out there for feedback.
How about `formats`?

It seems simple and right to the point and we know the dataset is coming from the `formats` configs: http://foremast.readthedocs.io/en/latest/configuration_files/foremast_config.html#formats.

jvasallo: Keeps it consistent. If I search `formats`, ill see it came from `utils` which is good.

src/foremast/securitygroup/create_securitygroup.py
e4r7hbug: I like exposing the Instance Attribute in case someone wants to override.

PR #314, opened 2018-07-25
Feature: Send additional data to create_app templates

sijis:
This passes `DEFAULT_RUN_AS_USER` and `formats` to `app_data.json.j2` template, which is used when creating an app.


jvasalloupgrade:
Looks good!

jvasallo:
Looks good from my side. Test and basically passing run_user down. Cool!

PR #315, opened 2018-07-26
fix: Expose Generator to User Data

e4r7hbug:
Need a way to access the standard name generator in User Data.

### DIFF
--- a/src/foremast/pipeline/construct_pipeline_block.py
+++ b/src/foremast/pipeline/construct_pipeline_block.py
@@ -141,7 +141,8 @@ def construct_pipeline_block(env='',
             env=env,
             region=region,
             project=generated.project,
-            region_subnets=region_subnets)
+            region_subnets=region_subnets,
+        )
     else:
         data = copy.deepcopy(settings)
 
@@ -181,7 +182,12 @@ def ec2_pipeline_setup(generated=None, project='', settings=None, env='', region
 
     """
     data = copy.deepcopy(settings)
-    user_data = generate_encoded_user_data(env=env, region=region, app_name=generated.app_name(), group_name=project)
+    user_data = generate_encoded_user_data(
+        env=env,
+        region=region,
+        generated=generated,
+        group_name=project,
+    )
 
     # Use different variable to keep template simple
     instance_security_groups = sorted(DEFAULT_EC2_SECURITYGROUPS[env])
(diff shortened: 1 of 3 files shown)


src/foremast/pipeline/construct_pipeline_block.py
```diff
@@ -179,14 +179,19 @@ def ec2_pipeline_setup(appname='', project='', settings=None, env='', region='',
...
+        env=env,
+        region=region,
+        generated=generated,
+        group_name=project,
+    )
 
     # Use different variable to keep template simple
     instance_security_groups = sorted(DEFAULT_EC2_SECURITYGROUPS[env])
-    instance_security_groups.append(appname)
+    instance_security_groups.append(generated.app_name())
```
sijis: Do you think there is a way to change this to another `Generator()` object? This is for cases where the default group name doesn't match the appname.

src/foremast/utils/encoding.py
```diff
@@ -16,16 +16,18 @@
 """Generate base64 encoded User Data."""
 import base64
 
+from gogoutils import Generator
```
sijis: Is this being used?

e4r7hbug: Nope. Thanks for catching.

src/foremast/pipeline/construct_pipeline_block.py
e4r7hbug: Where would this different Group name come from?

savilesupgrade: It would come from `gogoutilities.Generator()`. We could add another facility for `elb` and default it to the format used for `appname`. Just a thought.

sijis: It would come from `gogoutilities.Generator()`. We could add another facility for `elb` and default it to the format used for `appname`. Just a thought.

e4r7hbug: It sounds like you want `gogoutils.Generator.security_group` and `gogoutils.Generator.elb`. This could be achieved with the following and avoid another Instance of `Generator`.
```python
class Generator:
    def __init__(self):
        self.data = {
            'elb': '{app}',
            'security_group': '{app}',
        }


Generator(formats={
    'elb': 'custom_elb_name',
    'security_group': 'custom_name',
})
```

sijis: The changes that i was thinking were something like this: https://github.com/foremast/gogo-utils/pull/17

e4r7hbug: Updated from foremast/gogo-utils#17.

sijis:
All good now.

PR #319, opened 2018-08-16
Adding support for passing custom task uri

jvasallo:
Instead of rewriting a custom task module, I wish to update the existing task module to support passing of additional custom URI strings to tasks endpoint. This maintains feature parity but enables overriding custom tasks.


jvasallo:
Failing tests are hopefully resolved soon:
https://github.com/pytest-dev/pytest/issues/3774

PR #320, opened 2018-08-23
feat: Allow Wrapper passthrough keyword arguments

e4r7hbug:
This is useful for supporting custom variables being passed to the
Pipeline Wrapper Template.


savilesupgrade:
Looks like a straight forward change.
The tests seems to yell about the `render_template()` signature and I'm not sure why that is a syntax error in py34 though.

PR #321, opened 2018-09-18
Support yaml as application and pipeline config files

sijis:
This allows yaml files as valid configuration files for `pipeline` and `application-master-$env`.

### DIFF
--- a/src/foremast/utils/lookups.py
+++ b/src/foremast/utils/lookups.py
@@ -20,6 +20,7 @@ import os
 from base64 import b64decode
 
 import gitlab
+import yaml
 import requests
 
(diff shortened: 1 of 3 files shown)


src/foremast/utils/lookups.py
```diff
@@ -242,9 +243,9 @@ def json(self, branch='master', filename=''):
         file_contents = self.get(branch=branch, filename=filename)
 
         try:
-            json_dict = json.loads(file_contents)
+            json_dict = yaml.load(file_contents)
         # TODO: Use json.JSONDecodeError when Python 3.4 has been deprecated
```
bazimov: Should we remove this TODO since you are not using json lib anymore?

sijis: Probably a good idea. :+1:

PR #322, opened 2018-09-18
feat: Initial Plugin infrastructure for Application creation

e4r7hbug:
@sijis has a fairly stable Branch going for the new Plugin system. This should be a relatively safe switch over.

### DIFF
--- a/src/foremast/app/aws.py
+++ b/src/foremast/app/aws.py
@@ -1,0 +1,50 @@
+"""AWS Spinnaker Application."""
+import logging
+from pprint import pformat
+from foremast.app.base import BaseApp
+from foremast.utils import wait_for_task
+
+
+class SpinnakerApp(BaseApp):
+    """Create AWS Spinnaker Application."""
+
+    def __init__(self, pipeline_config=None, app=None, email=None, project=None, repo=None):
+        """Class to manage and create Spinnaker applications
+
+        Args:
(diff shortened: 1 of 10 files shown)


sijis:
@jvasallo @sbasgall @darinpantley Do you have any concerns/issues with merging this in?

It should not affect any app creation process besides adding another parameter named 'provider'.

src/foremast/app/aws.py
```diff
@@ -0,0 +1,35 @@
+"""AWS Spinnaker Application."""
+from pprint import pformat
+
+from foremast.app import base
+from foremast.utils import wait_for_task
+
+
+class SpinnakerApp(base.BaseApp):
+    """Create AWS Spinnaker Application."""
```
jvasallo: Is this still true or can provider be modified to support various apps

jvasallo:
I guess I am ok with the Pull Request, however, what is the end goal of this? Is this a way to extend foremast allowing foremast to be extended via various plugins and such (for example support for k8s, gcp)?

savilesupgrade:
Rebased by popular demand. :)

jvasallo:
Looks good to me. This is a major change so lets make this the dawn of the 4.x release.
@foremast/gogo @foremast/core @foremast/upgrade @foremast/redbox FYI

PR #323, opened 2018-09-20
feat: Allow developers to disable manual checkpoints before non-prod stages.

JessicaAlan:
Added and documented a new `pipeline.json` parameter:

`checkpoints`: List of accounts that will have a manual checkpoint stage before them. Order does not matter but these must correspond to the list passed to `env` above. If this option is not present, Foremast places manual checkpoints between all stages by default. A manual checkpoint will never be placed before the first stage. This parameter **does not** affect stages that are built using the ``*_prod_*.json.j2`` templates, which will always contain a manual checkpoint.

### DIFF
--- a/_docs/configuration_files/pipeline_json/index.rst
+++ b/_docs/configuration_files/pipeline_json/index.rst
@@ -85,6 +85,14 @@ List of accounts that the application will be deployed to. Order matters as it d
     | *Type*: array
     | *Default*: ``["stage", "prod"]``
 
+``checkpoints``
+~~~~~~~~
+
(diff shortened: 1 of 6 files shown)


jvasallo:
Looks like a safe contribution and welcomed contribution! Thanks!

I assume this is to enable development teams who want more control over manual judgement steps? Such that they could interject a manual judgement in a preprod environment?

JessicaAlan:
It's the other way around, but your idea is a more intuitive way of preserving backwards compatibility. I will change the name of the parameter to `skip_checkpoints` and update the logic and documentation accordingly.

PR #324, opened 2018-10-02
Allow developers to include or not include manual checkpoints

JessicaAlan:
using `checkpoints` param in pipeline.json. Defaults to old behavior if
param is missing.


JessicaAlan:
Can a maintainer take a look at this soon? It's a followup to #323

PR #327, opened 2019-02-04
Update pipeline_prod_s3.json.j2

dnava13:
Bug: Removed bad json line from internal template.


jvasallo:
Looks like a good fix to me! Thanks for your pull request! 👍

PR #328, opened 2019-02-11
Latest Spinnaker release returns additional results with missing keys

jvasallo:
Latest version of Spinnaker is returning additional subnets and VPCs. We have a strict conditional on specific key/values being present, and they might not always be present.

In addition, fixing issue with ast.literal_eval. Running into a reported bug after upgrading system Python to Python 3.7.2

### DIFF
--- a/src/foremast/consts.py
+++ b/src/foremast/consts.py
@@ -155,7 +155,7 @@ def _convert_string_to_native(value):
     result = None
 
     try:
-        result = ast.literal_eval(value)
+        result = ast.literal_eval(str(value))
     except (SyntaxError, ValueError):
         # Likely a string
         result = value.split(',')
--- a/src/foremast/utils/subnets.py
+++ b/src/foremast/utils/subnets.py
@@ -57,11 +57,10 @@ def get_subnets(
         raise SpinnakerTimeout(subnet_response.text)
 
     subnet_list = subnet_response.json()
-
     for subnet in subnet_list:
         LOG.debug('Subnet: %(account)s\t%(region)s\t%(target)s\t%(vpcId)s\t' '%(availabilityZone)s', subnet)
 
-        if subnet['target'] == target:
+        if 'target' in subnet and subnet['target'] == target:
             availability_zone = subnet['availabilityZone']
             account = subnet['account']
             subnet_region = subnet['region']
--- a/src/foremast/utils/subnets.py
+++ b/src/foremast/utils/subnets.py
@@ -60,7 +60,7 @@ def get_subnets(
     for subnet in subnet_list:
         LOG.debug('Subnet: %(account)s\t%(region)s\t%(target)s\t%(vpcId)s\t' '%(availabilityZone)s', subnet)
 
-        if 'target' in subnet and subnet['target'] == target:
+        if subnet.get('target', '') == target:
             availability_zone = subnet['availabilityZone']
             account = subnet['account']
             subnet_region = subnet['region']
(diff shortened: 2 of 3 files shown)


src/foremast/utils/subnets.py
```diff
@@ -57,11 +57,10 @@ def get_subnets(
         raise SpinnakerTimeout(subnet_response.text)
 
     subnet_list = subnet_response.json()
-
     for subnet in subnet_list:
         LOG.debug('Subnet: %(account)s\t%(region)s\t%(target)s\t%(vpcId)s\t' '%(availabilityZone)s', subnet)
 
-        if subnet['target'] == target:
+        if 'target' in subnet and subnet['target'] == target:
```
sijis: Wouldn't this work?
```suggestion
        if subnet.get('target', '') == target:
```

src/foremast/consts.py
```diff
@@ -155,7 +155,7 @@ def _convert_string_to_native(value):
     result = None
 
     try:
-        result = ast.literal_eval(value)
+        result = ast.literal_eval(str(value))
```
sijis: I'm wondering whats the value that got passed that requires this.

src/foremast/utils/vpc.py
```diff
@@ -51,7 +51,7 @@ def get_vpc_id(account, region):
 
     for vpc in vpcs:
         LOG.debug('VPC: %(name)s, %(account)s, %(region)s => %(id)s', vpc)
-        if all([vpc['name'] == 'vpc', vpc['account'] == account, vpc['region'] == region]):
+        if 'name' in vpc and all([vpc['name'] == 'vpc', vpc['account'] == account, vpc['region'] == region]):
```
sijis: `name` wasn't in the new output?
Are you concerned about the `account` or `region` missing?

jvasallo: Given how CloudDriver operates around Account and Region I think those are pretty hard set. The biggest issue is in foremast our vpc must be hardcoded named to vpc. I believe there is an issue with this.

After upgrading to the latest version of Spinnaker, I am seeing my default VPC in addition to my "vpc" where we do our operations. 

BTW, not a fan of my if check, if there is a more optimal let me know! 👍 Only did this as a quick fix. :)

src/foremast/consts.py
jvasallo: Sharing payload privately due to potential sensitivity.

src/foremast/utils/subnets.py
jvasallo: Good suggestion! :)

src/foremast/utils/vpc.py
```diff
@@ -41,7 +41,7 @@ def get_vpc_id(account, region):
             configured.
 
     """
-    url = '{0}/vpcs'.format(API_URL)
+    url = '{0}/networks/aws'.format(API_URL)
```
jvasallo: The old url was deprecated while digging:
https://github.com/spinnaker/gate/blame/a3f2416a330993878b715eaba96e50f11108fd7a/gate-web/src/main/groovy/com/netflix/spinnaker/gate/services/aws/InfrastructureService.groovy#L60

src/foremast/consts.py
sijis: No issue here after reviewing the scrubbed config.

src/foremast/utils/vpc.py
sijis: We already have an issue filed on the `vpc` naming issue https://github.com/foremast/foremast/issues/189.

sijis: I agree, this will work for now.

PR #329, opened 2019-02-24
feat: Adding support for AWS Secrets Manager

jvasallo:
Adding support for using AWS Secrets Manager.

### DIFF
--- a/src/foremast/templates/infrastructure/iam/kms.json.j2
+++ b/src/foremast/templates/infrastructure/iam/kms.json.j2
@@ -1,0 +1,15 @@
+        {
+            "Sid": "KMSDecrypt",
+            "Action": [
+                "kms:Decrypt"
+            ],
+            "Effect": "Allow",
+            "Resource": [
+            {%- for key_id in items %}
+                "arn:aws:kms:{{ region }}:{{ account_number }}:key/{{ key_id }}"
+                {%- if not loop.last -%}
+                ,
+                {%- endif -%}
+            {% endfor %}
+            ]
+        }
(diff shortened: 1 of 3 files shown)


src/foremast/templates/infrastructure/iam/kms.json.j2
```diff
@@ -0,0 +1,15 @@
+        {
+            "Sid": "KMSDecrypt",
+            "Action": [
+                "kms:Decrypt"
+            ],
+            "Effect": "Allow",
+            "Resource": [
+            {%- for key_id in items %}
```
sijis: Where is `items` coming from?

sijis:
How are some if these values, namely `items`, sent to these templates?

The templates do not seem to be included in any of the other ones.

Would adding these templates to your internal template files suffice?

src/foremast/templates/infrastructure/iam/kms.json.j2
jvasallo: Relying on the internal services code that creates items:
https://github.com/foremast/foremast/blob/432284ae32b7f30df99d1bb5eae0a393f9154dd1/src/foremast/iam/construct_policy.py#L111

Adding to our internal templates is fine, but just wanted to release a way of doing it to the community. :)

jvasallo: A good example of it used today is: https://github.com/foremast/foremast/blob/master/src/foremast/templates/infrastructure/iam/dynamodb.json.j2

PR #330, opened 2019-02-28
feat: Ability to pass repo_name and group_name to trigger

jvasallo:
Just adding the ability to pass the generated repo name to templates. Basically feature parity with other templates.


jvasallo:
I'll create an issue to maybe rework how we pass generated data downstream. Agreed will greatly simplify if utils ever adds more data. Change 1 function and all pipelines will get the benefit. 👍

PR #331, opened 2019-03-01
feat: Added support for custom runway directory name

jvasallo:
Not everyone will call their application repo "runway/", this enables folks to override this while still retaining the default "runway/" behavior today.


sijis:
All looks fine. 

The only thing really missing is are a couple of test cases validating the default value and a custom one.

PR #336, opened 2019-03-20
Refactor: Remove murl from creating urls

sijis:
Using murl strips out the `/path` part of a host. This becomes apparent under the following scenarios:

* Using a gate proxy
* Setting up gate under a subpath url
    
In these scenarios the host is specified as https://gate.url/path, then the calls to gate should be https://gate.url/path/api_endpoint, instead they become https://gate.url/api_endpoint.
    
This change allows those hosts definitions to work as expected.

This also remove `Gate()` class as references were removed in https://github.com/foremast/foremast/commit/51adda9e49cd4f1fc368494410bbac68f08640d6.


jvasallo:
@sijis This basically looks like what you see is what you get. We are removing our abstract URL merging to just allow users to specify their gate endpoint however it might be? For example: https://gate.example.com/<some_subdomain>/<...>
https://gate.example.bar/<...>
https://spinnakertest.com/<..>

jvasallo:
Verified locally against two endpoints and looks good. Tests passing as well. LGTM!

Thanks @sijis !

PR #340, opened 2019-07-11
Lambda Docs: Timeout Invalid

jvasallo:
The default timeout is not 3600, you can't even specify more than 900 currently. Added a doc in case AWS increases the timeout as well.


jvasallo:
Lint failures to be resolved in #339

PR #341, opened 2019-07-13
adding support to allow data pipeline in service block 

waynetaylor:
To have an event driven architecture its cool to use AWS lambda. When an S3 object is pushed today support is not available to activate a data pipeline. 

This new feature will add the capability.


jvasallo:
Looks good to me! Thanks for your first contribution to Foremast!

jvasallo:
Adding in 3.58.0 Release.

PR #342, opened 2019-07-29
fix: Added 'manual' pipeline type to default allowed types

linjmeyer:
Pipeline type "manual" is supported and included in documentation, however it is not in the default allowed pipeline types resulting in an exception when trying to use the feature.  This PR adds the manual pipeline type to the allowed pipelines.


jvasallo:
This is great! Documented feature but not enabled.

PR #345, opened 2019-08-07
WIP: Feature/kubernetes pipeline

linjmeyer:
First prototype of adding kubernetes support to Foremast.  I'm opening this draft for a code review/discussion on the changes made.  Due to fundamental differences between K8S and AWS the functionality of K8S pipelines is significantly different than AWS pipelines and should be discussed and validated with the community.

**Pipeline.json**
Pipeline.json file supports the following options:
```
"env": [ "spinnaker-k8s-account-1", "spinnaker-k8s-account-2"],
"type": "kubernetes",
"kubernetes": {
    "pipeline_type": "canary", // Determines the k8s template to be used.  Must match a template named pipeline_kubernetes_{type}.json.js
    "canaryAnalysisIntervalMins": 5, // optional, no default, the interval minutes of the ACA step.  Only needed if a kayenta ACA step is included in the template
    "canaryConfigName": "dotnet_core_http_500", // optional, no default, the name of the canary config to use.  Only needed if template includes a kayenta ACA step.
    "manifest_account_name": "spin-gcs-artifact-account" // optional, no default, the spinnaker account used to download manifests (e.g. from GCS)
}
```

**application-master-*.json**
Is not required or used during kubernetes deployments.  Generally speaking options defined here in AWS would be defined in the Kubernetes manifest used in the pipeline. (deployment, configmap, etc.)

**Pipeline changes**
Because Kubernetes pipelines, especially canaries, are much more flexible it is a difficult and error prone exercise to break the stages up into separate templates.  The many options in deploying using the Kubernetes v2 provider and difference in expected artifacts make it impossible to create practical reusable stages.  Due to this the following changes have been made to k8s pipelines:

* One pipeline is generated per environment listed
* Reusable/sharable pipeline stages are not supported
* Pipeline names use the format appname [environment] (was: appname [region])

**Canary configs**
* Canary configs must be defined outside of foremast
* Existing canaries can be specified by name.  The name will be translated to the canaryConfigId expected by Spinnaker automatically.
* Not all options in canary configs are supported by foremast yet, I will add these once we agree on the major changes listed above

**Other notable changes**
* Several method signatures expect region to be passed in, but the concept of a region in AWS does not exist within a K8S Cluster.  These methods have been reimplemented in the `SpinnakerPipelineKubernetesPipeline` class as needed. 

Remaining work (will be done if we can agree this is the correct way forward):
* K8S pipeline templates are using Redbox's internal templates and will be made more generic
* Additional canary options will be added (currently canary name and interval time is supported)
* Unit tests


src/foremast/pipeline/create_pipeline_kubernetes.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def get_canary_id(self, name):
+        """Finds a canary config ID matching the name passed.
+        Assumes the canary name is unique and the first match wins.
+        """
+        url = "{}/v2/canaryConfig".format(API_URL)
+        canary_response = requests.get(url, headers=self.header, verify=GATE_CA_BUNDLE, cert=GATE_CLIENT_CERT)
+
+        if not canary_response.ok:
+            raise SpinnakerPipelineCreationFailed('Pipeline for {0}: {1}'.format(self.app_name,
```
sijis: Are we sure that there is a json response (from `canary_response`) on a non-200 return code?

src/foremast/pipeline/create_pipeline_kubernetes.py
```diff
@@ -0,0 +1,196 @@
...
+                                                                                 canary_response.json()))
+
+        canary_configs = canary_response.json()
+        names = []
+        for config in canary_configs:
+            names.append(config['name'])
+            if config['name'] == name:
+                return config['id']
+
+        raise SpinnakerPipelineCreationFailed(
```
sijis: This could probably be added to the `for` loop before as an `else` condition, like so:
```
for i in range(100):
    names.append(i)
    if i == name:
        ...
else:
    raise SpinnakerPipelineCreationFailed(...)
```

src/foremast/pipeline/create_pipeline_kubernetes.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        url = "{}/v2/canaryConfig".format(API_URL)
+        canary_response = requests.get(url, headers=self.header, verify=GATE_CA_BUNDLE, cert=GATE_CLIENT_CERT)
+
+        if not canary_response.ok:
+            raise SpinnakerPipelineCreationFailed('Pipeline for {0}: {1}'.format(self.app_name,
+                                                                                 canary_response.json()))
+
+        canary_configs = canary_response.json()
+        names = []
```
sijis: Based on the usage, i'd probalby name this variable `canary_options`.

src/foremast/pipeline/create_pipeline_kubernetes.py
```diff
@@ -0,0 +1,196 @@
...
+        Args:
+        Returns:
+            bool: True if pipelines were created successfully
+        """
+
+        pipeline_envs = self.environments
+        self.log.debug('Envs from pipeline.json: %s', pipeline_envs)
+
+        deleted_pipelines = self.clean_kubernetes_pipelines(pipeline_envs)
+        self.log.debug('Deleted %s managed pipelines: %s', len(deleted_pipelines), deleted_pipelines)
```
sijis: This is likely better put in `clean_kubernetes_pipelines()`.
And if that's the case, there's no need to capture the output.

src/foremast/pipeline/create_pipeline_kubernetes.py
```diff
@@ -0,0 +1,196 @@
...
+                           k8s_pipeline_type, template_name)
+
+            pipeline_json = get_template(template_file=template_name, data=template_data, formats=self.generated)
+            pipeline_template = json.loads(pipeline_json)
+            # Merge template and wrapper into 1 pipeline
+            pipelines[env].update(pipeline_template)
+
+        self.log.debug('Assembled Pipelines:\n%s', pformat(pipelines))
+
+        for env, pipeline in pipelines.items():
```
sijis: If all you care is about the value, this could be written as:
```
if pipeline in pipeline.values():
   do_something(pipeline)
````

**Comment by sijis** [comment]
I'll think a little bit more about the changes.

The first thing that comes to mind is that these values below seem like they *would* belong in `application-master-*.json` config. The reason, is that that is env specific values. I would probably consider 'canary' to be env specific. I would probably want to change the options to not be camelCased, as the existing ones aren't that way.
```
"kubernetes": {
    "pipeline_type": "canary", // Determines the k8s template to be used.  Must match a template named pipeline_kubernetes_{type}.json.js
    "canaryAnalysisIntervalMins": 5, // optional, no default, the interval minutes of the ACA step.  Only needed if a kayenta ACA step is included in the template
    "canaryConfigName": "dotnet_core_http_500", // optional, no default, the name of the canary config to use.  Only needed if template includes a kayenta ACA step.
    "manifest_account_name": "spin-gcs-artifact-account" // optional, no default, the spinnaker account used to download manifests (e.g. from GCS)
}
```

> Reusable/sharable pipeline stages are not supported

Why not? Wouldn't that just make it a manual pipeline type? A quick look, i could see: canary, deploy, approval stages.

**Comment by linjmeyer** [comment]
@sijis I'll go through the code comments and update.  Between me not being a python dev and cut/pasting a lot to just get something functional together there's plenty of changes that need to be made.

> Reusable/sharable pipeline stages are not supported
>> Why not? Wouldn't that just make it a manual pipeline type? A quick look, i could see: canary, deploy, approval stages.

It all comes down to how flexible Kubernetes is and trying to create templates in Spinnaker involves a lot of SpEL, and your stages + artifacts are pretty tightly coupled.  Here is a very long winded example of things Foremast would need to account for when creating a pipeline:

What are you deploying: Istio configs, Deployment, ConfigMaps, etc.  Do you have it all in one manifest, broken out into 5 manifests (which needs 5 deploy steps) or do you want to bake them in Spinnaker using Helm (needs 1+ Bake steps dependent on the number of Helm Charts used).  Each deployment stage needs either inline YAML or the artifact ID which is defined in the J2 'wrapper' so Foremast needs to map that somehow.  If the manifests are external we need to provide a way to create them too and that brings up lots of new questions like where are they coming from: Jenkins, GCS, S3, etc.  How do you want each stage to fail?  It's not uncommon to have something like a cleanup step where the K8S provider deletes old configs, or deletes an immutable k8s object so the next deployment doesn't fail (Jobs for example).  If that's the case you need to set the deploys to not fail the pipeline and use SpEL in a later stage to Clean up objects in Kubernetes.  Do you plan on deploying to two clusters or one cluster separated by namespace, if so do you have multiple manifests or do you want Spinnaker to update the namespace value for you (using SpEL or Helm)? 

Everything listed is possible in a relatively simple setup, once you get into the Kayenta/ACA stuff it gets even worse.  Since Spinnaker+Kubernetes is sort of the wild-west without standardized best practices theres a crazy amount of options to account for.

You bring up a good point though: it is essentially the manual step + a function to resolve a canary ID by name.  Arguably we could extend the manual pipeline option with kubernetes targeted functionality.  I'd also add this approach is very similar to MPTv2 + Spin CLI with a few extra features.

File: src/foremast/pipeline/create_pipeline_datapipeline.py [hunk 6]
```diff
@@ -62,7 +62,7 @@ def render_wrapper(self, region='us-east-1'):
                 'base': base,
                 'deploy_type': deploy_type,
                 'environment': 'packaging',
-                'region': region,
+                'region': "region",
```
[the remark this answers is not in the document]
linjmeyer: Yes :)

src/foremast/templates/pipeline/pipeline_kubernetes_simple.json.j2
```diff
@@ -0,0 +1,252 @@
+{
+  "expectedArtifacts": [
+    {
+      "defaultArtifact": {
+        "id": "da60c92f-e404-40c6-8ace-8f0e60bb8f0f",
```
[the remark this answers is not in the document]
linjmeyer: Spinnaker expects the ID to be set and match the ID of the artifact, but on the Foremast side it could be possible to map them before sending it.  I'm doing something similar with the canary (accept a name, map to canary ID)

linjmeyer:
@sijis I'm closing the kubernetes PR.  Working on some upgrades to the manual pipeline type that will achieve everything we need, PR on that coming soon.  None of these changes are really kubernetes specific so it will fit better under the manual pipeline type.

PR #348, opened 2019-08-15
feat: Scheduled Actions for clusters

jvasallo:
This enables the ability to specify Scheduled Actions for Autoscaling groups to clusters. This is helpful in the event of shadow capacity for large known events such as email campaigns, Black Friday, and other business related scaling activities.


jvasallo:
This is 1/2 to support https://github.com/foremast/foremast/issues/338

PR #349, opened 2019-08-20
fix: don't overwrite s3 triggers set by other lambdas

JeffKo427:
**Problem:** Foremast uses the [put_bucket_notification_configuration](https://docs.aws.amazon.com/cli/latest/reference/s3api/put-bucket-notification-configuration.html) method of the AWS CLI to configure s3-based triggers for Lambda functions. This method does not upsert, but overwrites all triggers on the bucket.

**Solution:** Pull the current notification configuration and parse it for any triggers not set by this Lambda function. Copy those triggers into the new configuration.

**Caveat:** This may result in Foremast attempting to create two triggers with the same filter, causing an error.

**Possible Enhancement:** Copy the SQS and SNS triggers to ensure that they are preserved as well.

### DIFF
--- a/src/foremast/awslambda/s3_event/s3_event.py
+++ b/src/foremast/awslambda/s3_event/s3_event.py
@@ -64,6 +64,10 @@ def create_s3_event(app_name, env, region, bucket, triggers):
         for trigger in old_config["LambdaFunctionConfigurations"]:
             if trigger["LambdaFunctionArn"] != lambda_alias_arn:
                 config_json["LambdaFunctionConfigurations"].append(trigger)
+    if "QueueConfigurations" in old_config:
+        config_json["QueueConfigurations"] = old_config["QueueConfigurations"]
+    if "TopicConfigurations" in old_config:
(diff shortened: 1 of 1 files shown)


jvasallo:
We ran into this same issue, and we moved towards using S3 bucket notifications as a centralized source of managing bucket events vs the other way around to ensure state. 

S3 Pipeline with bucket notifications were used to let the bucket master determine when to trigger lambdas. https://foremast.readthedocs.io/en/latest/configuration_files/application_json.html#notification-keys

JessicaAlan:
I spoke with the team that requested this change; the bucket in question was not deployed through the pipeline and they see this change as the best solution for them. I can push back if you have concerns about this change.

waynetaylor:
@JessicaAlan  - I was just working on the same change :) - I just added a PR to let me take a look at yours vs mine

JessicaAlan:
Any update on this?

waynetaylor:
I solved mine with an s3 pipeline so will close out my items. I see you solved the main concern I had with yours to add sqs and sns listeners back as well.

Some thought does not to be done about management of removal

PR #350, opened 2019-08-23
Adding support to add multiple s3 notifications on same bucket

waynetaylor:
Appreciate some feedback on this to help solve the current limitation of one S3 event per bucket in current implementation


waynetaylor:
Closing out request. 

I've since contributed code to solve via an s3 pipeline which gives better overall management of the s3 bucket from a policy and notifications perspective.

If users want to continue there is a great example from Jessica

PR #353, opened 2019-08-29
feat: add tags to ELB after creation

JeffKo427:
Since Spinnaker does not currently support tagging ELBs for some reason, Gogo has been accomplishing this by having all new instances update their ELB's tags via an API call. We would prefer that these tags be added by Foremast in order to reduce the number of tools we are required to maintain.

This change is not expected to be breaking as any unwanted tags can be overwritten or simply not used, but we are opening to making this behavior optional via a parameter.

### DIFF
--- a/src/foremast/elb/create_elb.py
+++ b/src/foremast/elb/create_elb.py
@@ -14,6 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 """Create ELBs for Spinnaker Pipelines."""
+import os
 import json
 import logging
 from pprint import pformat
@@ -43,6 +44,7 @@ class SpinnakerELB:
         self.env = env
         self.region = region
         self.properties = get_properties(properties_file=prop_path, env=self.env, region=self.region)
+        self.git_commit = get_properties(properties_file=prop_path, env='pipeline')['config_commit']
 
     def make_elb_json(self):
         """Render the JSON template with arguments.
@@ -125,6 +127,7 @@ class SpinnakerELB:
         self.add_backend_policy(json_data)
 
         self.configure_attributes(json_data)
+        self.add_elb_tags()
 
     def add_listener_policy(self, json_data):
         """Attaches listerner policies to an ELB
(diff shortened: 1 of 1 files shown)


JessicaAlan:
Any feedback on this?

jvasallo:
This appears to be failing tests, was wondering if that was going to get fixed before review. Wasn't sure if this was completed.

src/foremast/elb/create_elb.py
```diff
@@ -274,3 +277,17 @@ def configure_attributes(self, json_data):
             LOG.debug('Load Balancer Attributes:\n%s', pformat(load_balancer_attributes))
             elbclient.modify_load_balancer_attributes(
                 LoadBalancerName=self.app, LoadBalancerAttributes=load_balancer_attributes)
+
+    def add_elb_tags(self):
```
jvasallo: Could this be moved to a utility function such as was done for s3?
https://github.com/foremast/foremast/blob/ee3b2290a6ad2d4a22c384237a17997e16a00d8b/src/foremast/s3/s3apps.py#L251

src/foremast/elb/create_elb.py
```diff
@@ -274,3 +277,17 @@ def configure_attributes(self, json_data):
...
+        elbclient = env.client('elb')
+
+        group = os.getenv("PROJECT")
+        repo = os.getenv("GIT_REPO")
+        tags = [{"Key": "Name",         "Value": '{}{}'.format(group, repo)},
+                {"Key": "app_name",     "Value": '{}{}'.format(group, repo)},
+                {"Key": "app_group",    "Value": group},
+                {"Key": "git_project",  "Value": '{}/{}'.format(group, repo)},
+                {"Key": "git_commit",   "Value": self.git_commit},
+                {"Key": "artifact_url", "Value": os.getenv("ARTIFACT_PATH")}]
```
jvasallo: In general, foremast typically only tags:
app_name and app_group 

I vote to remain consistent with that. Don't like pulling environment  (using os.getenv) variables in this fashion as they are not always guaranteed.

Could we limit tagging to app_name and app_group? These are already apart of foremast as well so no need to pull them from environment variables either.

jvasallo: With that being said, if you wish to extend it similar to S3 buckets implementation where users can specify additional tags, by all means! :)

src/foremast/elb/create_elb.py
```diff
@@ -14,6 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 """Create ELBs for Spinnaker Pipelines."""
+import os
```
jvasallo: mentioned below, but we should avoid getting environment variables this far into foremast unless really needed.

PR #354, opened 2019-09-03
feat: manual pipeline jinja template support

linjmeyer:
This PR is meant to address the feedback in [WIP: Feature/kubernetes pipeline #345](https://github.com/foremast/foremast/pull/345) where it was identified the same features/functionality could be added to the foremast generate pipeline manual type.

# New features for manual pipeline type
* Backwards compatibility with existing manual pipelines
* Support for Jinja template rendering (using .json.j2 file extensions)
* Allows defining variables in Jinja based manual pipelines which can be passed in via pipeline.json
* Ability to use pipeline templates defined in TEMPLATES_PATH (for sharing manual pipelines as templates, as opposed to storing the pipeline in the same repo/folder as the app)
* Utility function for resolving a CanaryConfigId from Spinnaker for the given canaryConfig name
* Ability for foremast to expose certain python functions into Jinja templates (used by `get_canary_id` function)
* Support for 'overloading' the manual pipeline type keyword in the foremast config.  For example, allowing `type: "kubernetes"` to trigger manual pipelines in addition to `type: "manual"`

The above features allow the manual pipeline step to operate as a Spinnaker pipeline templating engine which brings enough flexibility to support any sort of pipeline (kubernetes included).

## Example Pipeline Jinja Template

pipeline.json
```jsonc
{
    "deployment": "spinnaker",
    "type": "manual",
    "pipeline_files": [
        // templates:// tells foremast to find the file in the TEMPLATES_PATH defined in the foremast config
        // json.j2 means Jinja is used in the template and foremast should handle accordingly
        "templates://company-shared-template.json.j2", 
        // Treated as a local json file or in GitLab (same functionality as existing manual type)
       "my-local-pipeline.json"
    ],
    // Pipeline variables allow an array of key/value dictionaries to be passed to Jinja templates
    // Foremast does not evaluate or use the variables in anyway, just passes to the template
    // Variables should be defined as an array in the same order as the pipeline_files property
    "template_variables": [
        {
            "job_name": "a_lins_job",
        },
        {} // Empty value can be specified in the event multiple pipelines are defined but not all need variables
    ]
}
```

company-shared-template.json.j2 (just the relevant parts)
```jsonc
{
    "parameterConfig": [],
    "stages": [
        {
            "analysisType": "realTime",
            "canaryConfig": {
              // Calling the get_canary_id function which foremast exposes to the template
              "canaryConfigId": "{{ get_canary_id('my-canary-name') }}"
              // ...
            },
            // ...
          }
    ],
    "triggers": [
      {
        // Example using a template variable "job_name" which should be defined in pipeline.json's template_variables
        // All user-defined variables are accessible at template_variables.<var_name>
        "job": "{{ template_variables.job_name }}",
        // ...
      }
    ],
  }
```


src/foremast/consts.py
```diff
@@ -197,10 +197,15 @@ def _generate_security_groups(config_key):
 DOMAIN = validate_key_values(CONFIG, 'base', 'domain', default='example.com')
 ENVS = set(validate_key_values(CONFIG, 'base', 'envs', default='').split(','))
 REGIONS = set(validate_key_values(CONFIG, 'base', 'regions', default='').split(','))
+MANUAL_PIPELINE_SYNONYMS = set(
```
jvasallo: I don't like SYNONYMS. 😄 

Can we just make this a list of `MANUAL_PIPELINES`

Include the name "manual" in the default list.

src/foremast/runner.py
```diff
@@ -111,7 +111,7 @@ def create_pipeline(self, onetime=None):
                 spinnakerpipeline = pipeline.SpinnakerPipelineS3(**kwargs)
             elif pipeline_type == 'datapipeline':
                 spinnakerpipeline = pipeline.SpinnakerPipelineDataPipeline(**kwargs)
-            elif pipeline_type == 'manual':
+            elif pipeline_type == 'manual' or pipeline_type in consts.MANUAL_PIPELINE_SYNONYMS:
```
jvasallo: Could we simplify this to just:

`elif pipeline_type in consts.MANUAL_PIPELINES` and put the value of `"manual"` in the list?

src/foremast/consts.py
```diff
@@ -197,10 +197,15 @@ def _generate_security_groups(config_key):
...
+    validate_key_values(CONFIG, 'base', 'manual_pipeline_synonyms', default='manual').split(','))
 ALLOWED_TYPES = set(
-    validate_key_values(CONFIG, 'base', 'types', default='ec2,lambda,s3,datapipeline,rolling,manual').split(','))
+    validate_key_values(CONFIG, 'base', 'types', default='ec2,lambda,s3,datapipeline,rolling,manual')
```
jvasallo: Don't think we need to append `ALLOWED_TYPES` here. Since we are extending `manual` only with this PR, think we can assume if users allow the use of `"manual"` pipelines, they will also get access to various names.

We can do the check of manual pipelines names later in the logic.

src/foremast/utils/pipelines.py
```diff
@@ -115,3 +115,36 @@ def normalize_pipeline_name(name=''):
     for bad in '\\/?%#':
         normalized_name = normalized_name.replace(bad, '_')
     return normalized_name
+
+
```
jvasallo: General placement of the util. 

Would you consider this a pipeline utility or kayenta/canary util?

src/foremast/pipeline/jinja_functions.py
```diff
@@ -0,0 +1,26 @@
+#   Foremast - Pipeline Tooling
+#
+#   Copyright 2019 Gogo, LLC
```
jvasallo: Fix Copyright 👍

src/foremast/pipeline/create_pipeline_manual.py
```diff
@@ -14,29 +14,106 @@
...
+            file_name (str): pipeline file name
+
+        Returns:
+            str: Contents of pipeline file."""
+        # Check if this file is a shared template in the TEMPLATES_PATH
+        if file_name.startswith(TEMPLATES_SCHEME_IDENTIFIER):
+            file_name = file_name.lstrip(TEMPLATES_SCHEME_IDENTIFIER)
+            pipeline_templates_path = TEMPLATES_PATH.rstrip("/") + "/pipeline"
+            lookup = FileLookup(git_short=None, runway_dir=pipeline_templates_path)
+            return lookup.get(filename=file_name)
```
jvasallo: Can we add an explicit `else` and just have a single return?

src/foremast/consts.py
```diff
@@ -197,10 +197,15 @@ def _generate_security_groups(config_key):
...
+    .split(',')) | MANUAL_PIPELINE_SYNONYMS
 RUNWAY_BASE_PATH = validate_key_values(CONFIG, 'base', 'runway_base_path', default='runway')
 TEMPLATES_PATH = validate_key_values(CONFIG, 'base', 'templates_path')
+TEMPLATES_SCHEME_IDENTIFIER = validate_key_values(CONFIG, 'base', 'templates_scheme_identifier', default='templates://')
```
jvasallo: Naming, could we call this `TEMPLATES_URI_SCHEME`

src/foremast/consts.py
```diff
@@ -197,10 +197,15 @@ def _generate_security_groups(config_key):
...
+TEMPLATE_VARIABLES_KEY = validate_key_values(CONFIG, 'base', 'template_variables_key', default="template_variables")
```
jvasallo: Does this need to be configurable or could it just be a common template key? We can add it to the docs and that can help remove some of the logic checks you are doing as well?

src/foremast/consts.py
jvasallo: Call this `ALLOWED_MANUAL_TYPES`

src/foremast/consts.py
jvasallo: Remove the `|` and we can do the logic later.

src/foremast/runner.py
jvasallo: In an earlier check do an and check for `ALLOWED_MANUAL_TYPES`

src/foremast/utils/kayenta.py
```diff
@@ -0,0 +1,53 @@
+#   Foremast - Pipeline Tooling
+#
+#   Copyright Redbox Automated Retail, LLC
```
jvasallo: Year please!

jvasallo:
Looks good, Docs are slated to come with #351 as part of that change! 👍 🌮

PR #356, opened 2019-09-11
fix: Error when getting task status

sijis:
Sending the content-type header seems to cause a 400 bad request.
Removing this should not cause a problem as we are doing a GET request.

### DIFF
--- a/src/foremast/utils/tasks.py
+++ b/src/foremast/utils/tasks.py
@@ -13,6 +13,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 """POST a new task or check status of running task."""
+import copy
 import json
 import logging
 from functools import partial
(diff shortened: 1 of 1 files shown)


jvasallo:
@sijis I tested this, and it seems to work. Since this is the only place `HEADERS` is used, I would opt just to remove it from the `consts.py` 

When I do a `git blame`, I do see a legacy reference. Going to ask for some assistance to see why this was merged in the first place. If all is well, we should be able to add tomorrow morning and cut a release!

src/foremast/utils/tasks.py
```diff
@@ -75,8 +76,11 @@ def _check_task(taskid):
 
     LOG.info('Checking taskid %s', taskid)
 
+    headers = copy.copy(HEADERS)
```
jvasallo: Since HEADERS is only used here oddly enough, we should just remove it from consts.py. 👍

PR #359, opened 2019-09-18
Fixed bug causing incorrect config py to be loaded

linjmeyer:
When setting `FOREMAST_CONFIG_FILE=config2.py` I noticed it was still always loading `config.py`.  

The code to import the 'config' module was adding the config2.py's directory then loading the module, but since both config2.py and config.py are in the same directory (in my case) the incorrect one was loaded.  

This PR fixes the bug by loading the config module from a filepath rather than a directory.

The solution only works in Python 3.5+ according to [StackOverflow](https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path).


jvasallo:
Looks safe to me, we should still document this. Ill take care of this in a separate MR.

PR #361, opened 2019-10-03
refactor: Centralizing gate calls to a single function

jvasallo:


src/foremast/utils/gate.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def gate_request(method='GET', uri=None, headers=None, data=None, params=None):
+    """Make a request to Gate's API via various auth methods
+
+    Args:
+        method (str): Method to request Gate API; GET or POST
+        uri (str): URI path to gate API
+    """
+    response = None
```
sijis: You should do a `method = method.lower() #or method.upper()` to prevent silly issues like `gate_request(method='Post'...)` not working.

src/foremast/utils/gate.py
```diff
@@ -0,0 +1,50 @@
...
+
+    url = '{host}{uri}'.format(host=API_URL, uri=uri)
+
+    if OAUTH_ENABLED:
+        headers['Bearer'] = ""
+        raise NotImplementedError
+
```
sijis: No docs anywhere noted on this.

jvasallo: @sijis a PR is coming, but we are still drawing this up. We wanted to add it to this commit, but decided against it. Working on this currently.

src/foremast/utils/gate.py
jvasallo: Good idea!

PR #362, opened 2019-10-03
Added util function for Google Identity Aware Proxy tokens

linjmeyer:
Google Cloud supports a partially automated deployment of Spinnaker using Google Identity Aware Proxy for authentication.   There is only one Gate endpoint available, making it difficult to support Google IAP and x509 Auth.

This PR adds a util function to authenticate against Google IAP using an appropriately configured Google Cloud Service Account.  The function returns the a JWT which can be used to authenticate against Spinnaker.

Usage:

```python
svc_account_creds_path='/my/path/svc-account.json'
iap_client_id ='my-client-id-from-google'
my_token = get_google_iap_bearer_token(iap_client_id, svc_account_creds_path)
```

### DIFF
--- a/src/foremast/utils/google_iap.py
+++ b/src/foremast/utils/google_iap.py
@@ -2,16 +2,15 @@ import google.auth
 import google.auth.app_engine
 import google.auth.compute_engine.credentials
 import google.auth.iam
-from google.auth.transport.requests import Request
+from google.auth.transport.requests import Request as GoogleAuthRequest
 import google.oauth2.credentials
 import google.oauth2.service_account
-import requests
-import requests_toolbelt.adapters.appengine
 import logging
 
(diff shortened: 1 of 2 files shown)


src/foremast/utils/google_iap.py
```diff
@@ -0,0 +1,53 @@
...
+        service_account_credentials._make_authorization_grant_assertion())
+    # Get Google Signed JWT Token
+    request = GoogleAuthRequest()
+    body = {
+        'assertion': service_account_jwt,
+        'grant_type': google.oauth2._client._JWT_GRANT_TYPE,
+    }
+    token_response = google.oauth2._client._token_endpoint_request(
+        request, GOOGLE_OAUTH_TOKEN_URI, body)
+    return token_response['id_token']
```
jvasallo: Can we just return the full token_response?

jvasallo:
Looks good as a utility function, now that we are centralizing auth

PR #364, opened 2019-10-04
adding initial support for dlq

waynetaylor:
Currently foremast does not support Dead Letter Queue Configuration as part of Lambda Pipeline. 

A key challenge that Lambda users will face in the pipeline is the handling of exceptions. Some examples include:

- Notifications of when a function fails e.g. via an SNS handler
- Queue messages that can be reviewed by developers for root cause such as SQS


jvasallo:
Nice enhancement, should be good for those using SQS dead letters.

PR #366, opened 2019-10-04
fix: Unneeded loop of accounts on app creation

jvasallo:
Looking at the latest Spinnaker Release, the create application call does not expect an iterated loop of accounts. This is breaking some extensions when it comes to manual pipelines. In order to retain backwards compatibility we are leaving a default list of accounts but this isn't used. Here is the most recent JSON payload to createApplication:

```{
    "job": [
        {
            "type": "createApplication",
            "application": {
                "cloudProviders": "aws",
                "instancePort": 80,
                "providerSettings": {
                    "aws": {
                        "useAmiBlockDeviceMappings": false
                    }
                },
                "name": "testapp",
                "email": "test@test.com",
                "description": "Test"
            },
            "user": "example@example.com"
        }
    ],
    "application": "testapp",
    "description": "Create Application: testapp"
}
```

### DIFF
--- a/src/foremast/app/aws.py
+++ b/src/foremast/app/aws.py
@@ -17,7 +17,10 @@ class SpinnakerApp(base.BaseApp):
             AssertionError: Application creation failed.
 
         """
-        self.appinfo['accounts'] = self.get_accounts()
+
+        # Retaining abstract account list for backwards compatability
+        # Refer to #366
+        self.appinfo['accounts'] = ['default']
         self.log.debug('Pipeline Config\n%s', pformat(self.pipeline_config))
         self.log.debug('App info:\n%s', pformat(self.appinfo))
         jsondata = self.render_application_template()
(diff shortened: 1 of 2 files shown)


jvasallo:
@sijis we tested this quite a bit, and we think this might be some logic from when we initially thought each "app/environment" needed their own application. This was the old asgard logic if I recall where we had an app in each cluster.

jvasallo:
By leaving a default list, we ensure that folks with custom templates that might have forked off the base will still work, in addition we fix the weird dependency on AWS accounts to create apps. Not sure if this was some legacy Spinnaker thing, but the AWS dependency doesn't seem to be needed anymore. The old behaviour caused a weird loop where it would have 5 create app calls to Spinnaker in 1 json payload. It seemingly doesn't matter if we do one or many. 

As a result we are just doing one and removing the weird AWS dependency to begin work on other provider support.

PR #369, opened 2019-10-06
Fix: File imports

sijis:
Fix the isort warnings from https://github.com/foremast/foremast/pull/368.


jvasallo:
Nice! Thanks @sijis!

PR #375, opened 2019-10-07
Fixed asg index indicator

dnava13:
Fixed index format string from 0 to 1. Only index zero is available in the current context.


jvasallo:
Awesome fix. Thanks for catching this!

PR #381, opened 2019-11-14
incrementing the retry service

waynetaylor:
Was going to add https://pypi.org/project/backoff into code base - but found you had retry already implemented. 

I've increased retry count and timer across board. 

We may also want to consider using https://pypi.org/project/backoff as its wider adopted and also actively maintained.

### DIFF
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -152,9 +152,9 @@ class APIGateway:
             source_arn=global_api_source_arn)
     
     @backoff.on_exception(backoff.expo,
-                      botocore.exceptions.ClientError,
-                      max_tries=5,
-                      jitter=None)
+        botocore.exceptions.ClientError,
+        max_tries=5,
+        jitter=None)
     def create_api_deployment(self):
         """Create API deployment of ENV name."""
         try:
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -19,7 +19,7 @@ import logging
 
 import boto3
 import botocore
-from tryagain import retries
+import backoff
 
 from foremast.utils import (add_lambda_permissions, get_details, get_env_credential, get_lambda_alias_arn,
                             get_lambda_arn, get_properties)
@@ -150,8 +150,11 @@ class APIGateway:
             env=self.env,
             region=self.region,
             source_arn=global_api_source_arn)
-
-    @retries(max_attempts=10, wait=6, exceptions=(botocore.exceptions.ClientError))
+    
+    @backoff.on_exception(backoff.expo,
+                      botocore.exceptions.ClientError,
+                      max_tries=5,
+                      jitter=None)
     def create_api_deployment(self):
         """Create API deployment of ENV name."""
         try:
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -151,7 +151,7 @@ class APIGateway:
             region=self.region,
             source_arn=global_api_source_arn)
 
-    @retries(max_attempts=5, wait=2, exceptions=(botocore.exceptions.ClientError))
+    @retries(max_attempts=10, wait=6, exceptions=(botocore.exceptions.ClientError))
     def create_api_deployment(self):
         """Create API deployment of ENV name."""
         try:
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -150,7 +150,7 @@ class APIGateway:
             env=self.env,
             region=self.region,
             source_arn=global_api_source_arn)
-        
+
     @backoff.on_exception(backoff.expo,
                           botocore.exceptions.ClientError,
                           max_tries=5,
--- a/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
+++ b/src/foremast/awslambda/api_gateway_event/api_gateway_event.py
@@ -150,11 +150,11 @@ class APIGateway:
             env=self.env,
             region=self.region,
             source_arn=global_api_source_arn)
-    
+        
     @backoff.on_exception(backoff.expo,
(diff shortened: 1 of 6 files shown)


src/foremast/utils/elb.py
```diff
@@ -15,17 +15,17 @@
 #   limitations under the License.
 """Search for ELB DNS name."""
 import logging
+from tryagain import retries
```
jvasallo: Move this back as we lint our libraries together typically with isort.

src/foremast/utils/security_group.py
```diff
@@ -26,7 +26,7 @@
 LOG = logging.getLogger(__name__)
 
 
-@retries(max_attempts=5, wait=2, exceptions=(SpinnakerSecurityGroupError))
+@retries(max_attempts=10, wait=6, exceptions=(SpinnakerSecurityGroupError))
```
jvasallo: Let's leave this alone for now, this will result in potentially 10 API operations with a delay of 6 seconds; i.e. 60 seconds worst case. In addition to hitting Spinnaker's Gate API hard, this will also make subsequent AWS calls thus manifesting the problem higher.

As I understand the issue was around ELBs not populating in the API so let's target that for now.

src/foremast/utils/subnets.py
```diff
@@ -27,7 +27,7 @@
 
 
 # TODO: split up into get_az, and get_subnet_id
-@retries(max_attempts=6, wait=2.0, exceptions=SpinnakerTimeout)  # noqa
+@retries(max_attempts=10, wait=6.0, exceptions=SpinnakerTimeout)  # noqa
```
jvasallo: Same comment as security_group.py

src/foremast/awslambda/awslambda.py
```diff
@@ -211,7 +211,7 @@ def update_function_configuration(self, vpc_config):
 
         LOG.info("Successfully updated Lambda configuration.")
 
-    @retries(max_attempts=3, wait=1, exceptions=(SystemExit))
+    @retries(max_attempts=10, wait=3, exceptions=(SystemExit))
```
jvasallo: I think the max_attempts is off, but the wait should be increased to `3` as you have done. 1 second delay is potentially too aggressive.

src/foremast/awslambda/awslambda.py
```diff
@@ -164,7 +164,7 @@ def update_alias(self):
             LOG.info("Alias update failed. Retrying...")
             raise
 
-    @retries(max_attempts=3, wait=1, exceptions=(SystemExit))
+    @retries(max_attempts=10, wait=3, exceptions=(SystemExit))
```
jvasallo: I think the max_attempts is off and should be reverted back to `3` attempts for now, but the wait should be increased to 3 as you have done. 1 second delay is potentially too aggressive.

src/foremast/awslambda/awslambda.py
```diff
@@ -152,7 +152,7 @@ def create_alias(self):
             LOG.info("Alias creation failed. Retrying...")
             raise
 
-    @retries(max_attempts=3, wait=1, exceptions=(boto3.exceptions.botocore.exceptions.ClientError))
+    @retries(max_attempts=10, wait=3, exceptions=(boto3.exceptions.botocore.exceptions.ClientError))
```
jvasallo: I think the max_attempts is off and should be reverted back to 3 attempts for now, but the wait should be increased to 3 as you have done. 1 second delay is potentially too aggressive.

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
```diff
@@ -151,7 +151,7 @@ def add_permission(self):
             region=self.region,
             source_arn=global_api_source_arn)
 
-    @retries(max_attempts=5, wait=2, exceptions=(botocore.exceptions.ClientError))
+    @retries(max_attempts=10, wait=6, exceptions=(botocore.exceptions.ClientError))
```
jvasallo: same comment as security_group.py

src/foremast/utils/elb.py
```diff
@@ -15,17 +15,17 @@
...
-from tryagain import retries
 
 from ..exceptions import SpinnakerElbNotFound
 from ..utils.gate import gate_request
 
 LOG = logging.getLogger(__name__)
 
 
-@retries(max_attempts=5, wait=2, exceptions=(AssertionError, SpinnakerElbNotFound))
+@retries(max_attempts=6, wait=10, exceptions=(AssertionError, SpinnakerElbNotFound))
```
jvasallo: Lets leave max attempts the same, but increase the wait to 5 seconds.

waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

src/foremast/awslambda/api_gateway_event/api_gateway_event.py
waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

src/foremast/awslambda/awslambda.py
waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

src/foremast/awslambda/awslambda.py
waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

src/foremast/awslambda/awslambda.py
waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

src/foremast/utils/subnets.py
waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

src/foremast/utils/security_group.py
waynetaylor: I have replaced this with the backoff library per my initial recommendation in my pull request. I have changed the max attempts back to the original values per request

waynetaylor:
Any updates on when this will be accepted @jvasallo

Issue #382, opened 2019-11-18
Feature Request: Support for a collection of manual templates


linjmeyer:
With Kubernetes pipelines it's common for a deployment to be split into different pipelines where one pipeline finishing triggers another pipeline to start.  For example, after a canary pipeline finishes you may want to trigger a second pipeline to cleanup the canary deployment.

`pipeline.json` does support multiple manual templates, but if template A has a dependency on template B it would be up to the developer to include both. 

I see two ways to support this:
* Add support for a template collection which is referenced by devs in their pipeline.json file.  A template collection consists of 1+ templates
  * Might be easily implemented by supporting folders, where if a folder is referenced all template files in the folder are used
* Add support for template dependencies
  * `template-a.json.j2` somehow tells Foremast it has a dependency on `template-b.json.j2`
  * Would require adding additional fields to the templates for Foremast to use.  Currently they are strictly Spinnaker Pipeline json with no added properties

PR #389, opened 2019-12-20
Feature/multi pipeline manual templates

linjmeyer:
Implements #382 

A single manual pipeline template (both JSON and J2) can now output either:
* A single pipeline object 
`{ application: "my app", ... }`
* An array of pipeline objects 
`[ { application: "my app", ... }, { application: "my app 2", ... } ]` 

The primary use case is to support creation of pipelines that are dependent on each other.  For example, a canary deployment pipeline and a canary cleanup pipeline that triggers on completion of the canary deployment pipeline.  Use cases that require this sort of setup can generally be solved with SPEL but in my experience this is clunky and error prone.

### DIFF
--- a/_docs/configuration_files/advanced_usages/manual_pipelines.rst
+++ b/_docs/configuration_files/advanced_usages/manual_pipelines.rst
@@ -93,7 +93,7 @@ Getting a Spinnaker Pipeline's JSON Body
          
          .. code-block:: json
 
-            {
+            [{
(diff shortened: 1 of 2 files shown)


jvasallo:
Looks good. Only comment is that as we are expecting this to be a list in the long run, we should probably update docs to reflect that. Maybe add some flavor around it explaining in previous releases you can specify a string, but overall update examples to be a list for the future :)

PR #406, opened 2020-05-28
Update Travis Syntax

edumco:
### Update Travis Syntax to remove some deprecated or renamed commands

This PR changes the following commands:

- Adds `os` explicit call
- Removes deprecated `sudo` command
- Moves the cache command to after script to replicate [the order of cache.](https://docs.travis-ci.com/user/caching#build-phases)
- Moves notifications to the end since is the last action taken
- Renames deprecated `matrix` for `jobs`
- Renames deprecated `user` for `username`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
-dist: xenial
-sudo: false
 language: python
-cache: false
 
-matrix:
+os: linux
+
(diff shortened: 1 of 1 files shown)


jvasallo:
Thanks for your contribution!

PR #413, opened 2020-07-29
feat: GCP Infrastructure and IAM support

linjmeyer:
Marking as draft for now, need to finish docs and linting once we agree on the way forward.

-------

Adds support for Google Cloud Platform (GCP) Identity and Access Management (IAM)

Breaking Changes:
* Foremast configuration options `MANUAL_TYPES` and `ALLOWED_TYPES` have been split into `AWS_MANUAL_TYPES`, `AWS_TYPES`, `GCP_MANUAL_TYPES` and `GCP_TYPES` to allow Foremast to invoke the correct cloud infrastructure functions

New features:
* Added support for defining a GCP environment as a collection of GCP Projects
* Added support for creating GCP Service Accounts
* Added support for project level IAM role bindings for service accounts

## Defining GCP Environments

GCP does not provide a way to define environments like stage, prod, etc.  Instead a GCP organization can create 1 or more projects, and it is up to the organization administrators to control permissions between projects as needed.  

Foremast uses the GCP APIs to query all projects which are labeled with `cloud_env=<env_name>`.  Projects in GCP can be labeled in the Cloud Console and using the gcloud CLI.  This provides a flexible way to define environments without restricting Foremast user's to a certain project naming convention or structure.

Foremast administrators define GCP environments in config.py or foremast.cfg like so:

```
'gcp': {
        'envs': {
            'dev': {
                'organization': 'my-organization-in-gcp.com',
                'service_account_project': 'my-project-where-svc-accounts-are-created',
                'service_account_path': '/path/to/service/account/for/foremast.json'
            }
        },
    }
```

Foremast will then find all projects in this environment by querying the GCP Project API for projects with the label `cloud_env=dev`.

## Defining IAM permissions/roles in pipeline.json

Similar to AWS, permissions are defined in the pipeline.json file's services block.  Roles requested by the repository are done project-by-project like so:

```
{
    "deployment": "spinnaker",
    "type": "manual",
    "env": [
      "stage",
      "prod"
    ],
    "services": {
        "gcp_roles": [
            {
                "project_prefix": "my-project",
                "roles": [
                    "roles/secretmanager.secretAccessor"
                ]
            }
        ]
    }
}
```

The `"project_prefix": "my-project",` is not the actual project name, but instead the prefix of the project.  The real project could be `my-project`, `my-project-stage` and `my-project-prod` or any combination of these.  In the future Foremast could support wildcards, regex or even determining the correct project for each environment using label/values.

## How Foremast updates policies in GCP

Permissions in GCP are defined in policies.  Policies have a list of bindings, which link accounts and roles.  For example, the IAM policy for project-a could container a binding for svcaccount1 and role/secretsmanager.secretsAccesssor.  This poses a challenge for Foremast however, because policies are shared per project.  New policies cannot be created, and GCP APIs do not offer a partial or "upsert" method to update policies.

Foremast gets a copy of all project IAM policies for a given environment and determines if bindings need to be removed/created for the repository's service account.  If changes are made, the entire policy must be updated locally and sent to GCP.  It is critical that Foremast does not update policy bindings for service accounts unrelated to the current repository.

If a 409 Conflict is returned when saving policies there is retry logic with a simple backoff algorithm to reduce the number of GCP API calls as much as possible.  409 Conflict is returned by GCP APIs when a policy is updated more than once in a short amount of time.


src/foremast/utils/gcp_service_account.py
```diff
@@ -0,0 +1,27 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Retrieve GCP Credentials"""
+
+from google.oauth2 import service_account
+from ..utils.gcp_environment import GcpEnvironment
+
+
+def get_credentials(env: GcpEnvironment):
```
jvasallo, line 22: What is the difference between this get_creds call vs the built in one on the class? Do they need to be separate?

src/foremast/gcp_iam/create_iam_resources.py
```diff
@@ -0,0 +1,88 @@
...
+    # For each project in this GCP environment, get the policy
+    # remove any references to this svc account, then re-add based on pipeline.json
+    # If changes were made locally, update the policy in GCP
+    for project in env.get_all_projects():
+        # Policies in GCP are shared between users/service accounts
+        # Attempt to update this project policy, if GCP returns a
+        # a 409 conflict we must re-retrieve the policy, re-update it
+        # and attempt to save again to save it again
+        attempts = 0
+        max_attempts = 5
```
jvasallo, line 32: https://github.com/foremast/foremast/blob/da668786740a9247c4fb7d77f3c651ef6084dd7e/src/foremast/utils/elb.py#L28

might be better to use the existing retries decorator on a function call? Thoughts?

src/foremast/gcp_iam/service_account.py
```diff
@@ -0,0 +1,64 @@
...
+"""GCP Service Account helpers"""
+
+import logging
+
+import googleapiclient.discovery
+
+LOG = logging.getLogger(__name__)
+
+
+def create_service_account(credentials, project_id, name):
```
jvasallo, line 25: This is creating the unique service account for mapping app to account right?

Foobar app == foobar service account

src/foremast/runner.py
```diff
@@ -130,11 +132,19 @@ def create_pipeline(self, onetime=None):
...
 
-    def create_iam(self):
-        """Create IAM resources."""
-        utils.banner("Creating IAM")
+    def create_aws_iam(self):
+        """Create AWS IAM resources."""
+        utils.banner("Creating AWS IAM")
         iam.create_iam_resources(env=self.env, app=self.app)
 
+    def create_gcp_iam(self, env: GcpEnvironment):
```
jvasallo, line 140: Assuming split was made with some thought. No ability under iam to have gcp and aws. if not in this release, ability to refactor in the future? Not sure if much code similarities or not between AWS/GCP.

jvasallo:
Any of the above captured in _docs? :) Lots of good details and effort here, but if I had to learn where to get started we should add maybe even a dedicated section to GCP IAM. Maybe even explain differences and such too.

jvasallo:
Another comment around:
```
The "project_prefix": "my-project", is not the actual project name, but instead the prefix of the project. The real project could be my-project, my-project-stage and my-project-prod or any combination of these. In the future Foremast could support wildcards, regex or even determining the correct project for each environment using label/values.
```

It might be easier to support wildcards, or somehow figure out a way to limit scope here. I don't think a manual mapping is feasible but would be nice to have some way to say, theses apps (or group) have access to these folders/projects. Essentially preventing creep outside the folders. :)

src/foremast/runner.py
linjmeyer: The implementations are very different for sure, but if there were a pipeline type that needed both it wouldn't be hard to invoke both sides of Foremast.  There would be a bit more to think about like, do environments in AWS match the names in GCP and things like that but definitely doable

src/foremast/gcp_iam/service_account.py
linjmeyer: Yes, the service account is made in the service account project defined in the GCP Environment in Foremcast config.  The service account name is the app name.  So say the project is coolapp/greatteam you'd end up with:
coolappgreatteam@prod-project-name.iam.gserviceaccount.com, coolappgreatteam@stage-project-name.iam.gserviceaccount.com, etc.

linjmeyer:
For the project_prefix, I have no issue with wildcards instead.  Another easy way in GCP APIs is to use the labels you can add to projects.  So it's very easy to do something like "get me the project where cloud_env=prod and team_name=myteam" or whatever other arbitrary labels.

For folders - that was one of my first approaches but the GCP Projects API only returns the parent folder id which is a random int.  Then the folders API doesn't have any endpoints to get info on that folder (e.g. the friendly name), just policy documents for the folder.   At least not that I can find.  In addition to all that, you have to query every one individually.  Similar to IAM, there is no "give me the structure of the entire org" style endpoint, just lots of individual API calls.

Foremast can only query projects it is given permissions to, plus those projects need `cloud_env` label on them, and then cloud_env value has to match the env in the pipeline.json file.

src/foremast/gcp_iam/create_iam_resources.py
linjmeyer: Yep done, I think a backoff is necessary and thought that wasn't possible with this library but I was wrong.  Fixed!

src/foremast/utils/gcp_service_account.py
linjmeyer: Left over from a code refactor, this function wasn't used at all.  Fixed!

src/foremast/gcp_iam/create_iam_resources.py
jvasallo: Awesome. Thanks. Looks much nicer.

src/foremast/utils/gcp_service_account.py
jvasallo: Awesome. Thanks.

jvasallo:
> Another comment around:
> 
> ```
> The "project_prefix": "my-project", is not the actual project name, but instead the prefix of the project. The real project could be my-project, my-project-stage and my-project-prod or any combination of these. In the future Foremast could support wildcards, regex or even determining the correct project for each environment using label/values.
> ```
> 
> It might be easier to support wildcards, or somehow figure out a way to limit scope here. I don't think a manual mapping is feasible but would be nice to have some way to say, theses apps (or group) have access to these folders/projects. Essentially preventing creep outside the folders. :) Alternatively, provide a way for admins to restrict specific projects/folders might be sufficient too.

For this, we discussed using foremast_enabled label on the GCP Project. This makes sure that only projects enabled in the GCP Console with the labels cloud_env and foremast_enabled are supported for the pipeline. This gives Administrators an ability to limit projects.

Another idea: foremast_group label to the APP_GROUP within foremast truly locks in applications to their own namespace. We are going to investigate if the API can be used to filter and compare GCP Label Group to Foremast internal APP_GROUP. If its not able to be done in the API, we might just need to return the result and evaluate in foremast.

So GCP labels:
CLOUD_ENV = prod
FOREMAST_ENABLED = true
FOREMAST_GROUPS = dataops,ondemand

jvasallo:
Pushing this for a 5.x Release. Labeling release.

src/foremast/runner.py
```diff
@@ -268,29 +278,60 @@ def cleanup(self):
...
-    runner.create_app()
 
+    try:
+        pipeline_type = runner.configs['pipeline']['type']
+        if pipeline_type is None or pipeline_type.isspace():
+            raise ForemastError("pipeline.type is required")
+    except KeyError:
+        raise ForemastError("pipeline.type is required")
+
+    if pipeline_type in consts.GCP_TYPES:
```
jvasallo, line 290: Maybe add backward compat logic.

If GCP_TYPES is null and AWS_TYPES is null flip to old config logic.

This saves our users who accidentally do major upgrades automatically from breaking changes :)

linjmeyer: If the older config options `types` and `manual_types` are set in the Foremast config, a deprecation warning is logged and those values are mapped to the new `AWS_*` equivalents. Commit: [b3aa70](https://github.com/foremast/foremast/pull/413/commits/b3aa70fbd097334ab29f4e490357c6c6f7b7007a)

linjmeyer:
As discussed the configuration changes in 5.x are now backwards compatible with older config files from <5.x.  A deprecation warning is logged asking that the new AWS_ and GCP_ equivalents are used.

Foremast now requires the following labels on GCP projects before making any changes to it's IAM policies:
```
CLOUD_ENV = {env_name}
FOREMAST_ENABLED = true
```

If it is desired that only certain groups be permitted to get roles on a given project, it can be set as a label on the GCP project like:
```
FOREMAST_GROUPS = team1,team2,team3
```

For example, this will prevent new roles from being granted on the given project for a group named team5

linjmeyer:
@jvasallo docs are done and discussed changes have been implemented!

src/foremast/utils/gcp_environment.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # No cached response found, check GCP APIs
+        service = discovery.build('cloudresourcemanager', 'v1', credentials=self.get_credentials())
+        project_filter = self._get_project_api_filter(name=project_name)
+        request = service.projects().list(filter=project_filter)
+        response = request.execute()
+        projects = response.get('projects', [])
+
+        # No projects found
+        if len(projects) == 0:
```
jvasallo, line 92: One small fix:
```
if not projects:
```

is technically the pythonic way of checking this.

src/foremast/consts.py
```diff
@@ -245,8 +240,49 @@ def _generate_security_groups(config_key):
...
+    LOG.warning("types is deprecated, use gcp_types and aws_types instead")
+    AWS_TYPES = set(_legacy_allowed_types.split(','))
+
+# Allowed and manual pipelines types for all cloud providers
+ALLOWED_TYPES = AWS_TYPES.union(GCP_TYPES)
+MANUAL_TYPES = AWS_MANUAL_TYPES.union(GCP_MANUAL_TYPES)
+
+# Check if types have accidentally been defined for both clouds
+_duplicate_allowed_types = AWS_TYPES.intersection(GCP_TYPES)
+if len(_duplicate_allowed_types) > 0:
```
jvasallo, line 280:
```
if _duplicate_allowed_types:
```

Small fix again. Same can be applied to manual_types on L285

PR #414, opened 2020-08-21
Feature/gcp print env

linjmeyer:
In Foremast, much of a GCP environment is defined in GCP itself and not in Foremast's config.  It would be helpful if Foremast could output of how it sees all GCP environments.  This could help identify permissions issues with Foremast's GCP service accounts, help in debugging and documentation.  For feature parity, printing AWS environments is supported but only provides a simple list from the foremast configuration.

Changes:
* Added feature to print a formatted table of GCP and AWS environments

Usage:
`foremast describe environments {gcp,aws}` will produce something like:

```
╒═══════════════╤═════════════════════════╤════════════════════╕
│ Environment   │ Project                 │ Permitted Groups   │
╞═══════════════╪═════════════════════════╪════════════════════╡
│ dev           │ my-project-one          │ N/A                │
├───────────────┼─────────────────────────┼────────────────────┤
│ dev           │ my-project-two          │ group1,group2      │
╘═══════════════╧═════════════════════════╧════════════════════╛
```

You can also change the output format, e.g. if you want to copy/paste into confluence:
`foremast infra print-environment --print-table-format=jira` will produce something like:

```
|| Environment   || Project                 || Permitted Groups   ||
| dev           | my-project-one | N/A                |
| dev           | my-project-two    | group1,group2            |
```

and `--print-to-file=myfile.txt` will output the table into myfile.txt.

Example of AWS output:

```
╒═══════════════╕
│ Environment   │
╞═══════════════╡
│ env1          │
├───────────────┤
│ env2          │
├───────────────┤
│ env3          │
├───────────────┤
│ env4          │
├───────────────┤
│ env5          │
╘═══════════════╛
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ google-cloud-secret-manager==0.2.0
 google-api-python-client==1.7.11
 google-auth==1.11.2
 google-auth-httplib2==0.0.3
+tabulate==0.8.7
--- a/src/foremast/__main__.py
+++ b/src/foremast/__main__.py
@@ -16,7 +16,7 @@ def add_infra(subparsers):
     """Infrastructure subcommands."""
     infra_parser = subparsers.add_parser('infra', help=runner.prepare_infrastructure.__doc__)
     infra_parser.set_defaults(func=runner.prepare_infrastructure)
-    infra_print_subparser = infra_parser.add_subparsers(title='Print GCP Environments')
+    infra_print_subparser = infra_parser.add_subparsers(title='Infra Subcommands')
     print_env_parser = infra_print_subparser.add_parser("print-environment", help=runner.print_gcp_environments.__doc__)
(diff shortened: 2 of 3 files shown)


linjmeyer:
Updated to use `foremast describe environments {gcp,aws}`.  This should be more extendable with other `foremast describe ...` options in the future.  It also now supports AWS, I think there is room for improvement on the AWS side but it would make more sense to create some sort of abstraction in Foremast for environments/cloud providers and have `GcpEnvironment` and `AwsEnvironment` classes that follow that abstraction before spending too much time on that.  That would make it much easier to support both clouds in similar ways.

jvasallo:
Yes in a later release we can add extraction from Spinnakers Env API to get the account info and such. Would probably be the best, but we can figure it out. Ill create a open issue to track it :)

jvasallo:
https://github.com/foremast/foremast/issues/416

PR #417, opened 2020-09-04
fix: bug stripping too many template path chars

linjmeyer:
It seems `str.lstrip("templates://")` is sometimes stripping an additional character.  For example `templates://pipline_stage.json` is being stripped to `ipline_stage.json` and causing pipeline failures.  

Not entirely sure why, but we already know the file path starts with `templates://` at this point in the code so the easy fix is to remove it using Python's substring syntax instead.


linjmeyer:
Merging myself as this is a small change that is blocking several pipelines internally

PR #421, opened 2020-09-08
fix: missing method error in aws lambda deploy

linjmeyer:


dnava13 (approved):
Verified correct function name

PR #423, opened 2020-09-18
fix: Existing svc accounts not being found in GCP IAM

linjmeyer:
Fixes issue where Foremast does not see existing service accounts.  Cause was pagination in the list service accounts API. 

Changes:
* GCP get service account uses get svc account API (instead of list svc accounts api)
* Disables cache discovery on all GCP API objects as this fills the log with warning level exceptions ([see issue](https://github.com/googleapis/google-api-python-client/issues/299))


marinkovicpetar (approved):
Looks good to me, and this approach seems more feasible

PR #428, opened 2020-10-07
feat: IAM access control for Cloud Functions

linjmeyer:
GCP Cloud Functions using HTTP Triggers support anonymous/unauthenticated access as well as authenticated access.  With authenticated access it is also possible to grant access to invoke the cloud function to specific groups, users and service accounts.  Currently Foremast does not support any of these features, so all functions default to authenticated users only who have correct permissions on the entire project the function is in.

This PR allows opting in to unauthenticated access, as well as adding IAM bindings to grant access to a single function.


jvasallo (approved):
Nice! Jon Snow will be proud

PR #434, opened 2020-10-22
feat: Added support to timeout approval stages after specified time

jvasallo:
Today, pipeline stages timeout after 72 hours. This is done to maintain a clear queue and also performance to prevent stale pipelines. Most pipelines get hung up on approval stages, and as a result sometimes clutter a deployment view.

This feature enables the ability to maintain cleaner pipelines and fail faster than default 72 hour timeout by specifying an approval stage timeout in milliseconds (based on API).

### DIFF
--- a/_docs/configuration_files/application_json.rst
+++ b/_docs/configuration_files/application_json.rst
@@ -36,6 +36,17 @@ Describes the application.
     | *Type*: string
     | *Default*: ``null``
 
+``approval_timeout``
+*******************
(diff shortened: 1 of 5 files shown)


Windfall890:
💸  LGTM 💸

Windfall890:
Thanks a bunch, this is going to clean up our Deployment view a lot, we can always restart a timed out pipeline if we missed it!

PR #438, opened 2020-10-27
feat: Pipeline notification support for all providers and options

linjmeyer:
Foremast supports Spinnaker pipeline failure notifications for Slack and Email via the `notifications` block in `pipeline.json`.  

Today Spinnaker has a long list of features not supported by Foremast including Microsoft Teams, Google Chat, Pub/Sub, custom messages and triggering based on pipelines starting, failing and completing.  The new alpha features for Spinnaker Plugins also allows custom notification options to be defined ([example](https://github.com/spinnaker-plugin-examples/notificationPlugin)).  This PR adds support for any notification option.

**Deprecation of notifications block**

To achieve support for any notification option a new syntax was needed which mirrors the Spinnaker syntax.  This new option can be defined in `pipeline.json` using the `pipeline_notifications` block, which deprecates the current `notifications` block.  To avoid breaking existing pipelines and custom templates `notifications` block still functions without any changes, but is deprecated in the documentation and should not be given any new features.

See doc changes in this PR for examples on new notification configuration.

### DIFF
--- a/src/foremast/templates/pipeline/pipeline_notifications.json.j2
+++ b/src/foremast/templates/pipeline/pipeline_notifications.json.j2
@@ -1,0 +1,27 @@
+[
+{#- v1: data.app.slack and data.app.email are legacy options, supported only for backwards compatibility #}
+{%- if data.app.slack or data.app.email %}
+    {
+      "level": "pipeline",
+      "when": [
+        "pipeline.failed"
+      ],
+      "type": "slack",
+      "address": "{{ data.app.slack }}"
+    },
+    {
+      "level": "pipeline",
+      "when": [
+        "pipeline.failed"
+      ],
+      "type": "email",
+      "address": "{{ data.app.email }}"
+    }
+{%- elif data.app.pipeline.pipeline_notifications %}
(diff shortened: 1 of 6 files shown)


src/foremast/templates/pipeline/pipeline_notifications.json.j2
```diff
@@ -0,0 +1,27 @@
...
+      ],
+      "type": "email",
+      "address": "{{ data.app.email }}"
+    }
+{%- elif data.app.pipeline.pipeline_notifications %}
+{#- v2 pipeline.pipeline_notification configuration block: #}
+    {%- for notification in data.app.pipeline.pipeline_notifications %}
+    {{ notification | tojson }}
+    {% if not loop.last %},{% endif %}
+    {%- endfor %}
```
linjmeyer, line 25: A note on this, I tried just calling `tojson` on the entire array, but it seems after python parses pipeline.json and feeds it back into Jinja, it's technically a dictionary, which was causing issues with `tojson`.  The for loop helped me get around that issue and ensure the output was a JSON array.

jvasallo:
Nice, this should add Microsoft Teams support! :)

PR #440, opened 2020-11-03
fix: GCP foremast_groups label now double underscore separated

linjmeyer:
Im not entirely sure if this has changed, or if I missed this in testing - but GCP Project labels do not allow commas.  This means it's not possible to support multiple groups in the `foremast_groups` label.  

This PR changes from a comma separated list to a double underscore separated list.  Chose double underscore as it seems uncommon enough that it won't overlap with anything else.

Old (won't work):
image

New:
image


jvasallo:
It seems this is the only solution for now, however isn't there only a 64 character limit on this variable?

PR #442, opened 2020-11-11
Feature/manual templates generate pipeline

linjmeyer:
**Changes:**
* Adds Jinja function for manual templates `generate_predictable_pipeline_id` which will create reproducible/predictable Spinnaker pipeline Ids for an application and pipeline name.  It must be explicitly called in a Foremast manual template, so this is backwards compatible and opt-in only.

**Summary:**

Ran into an interesting chicken-and-egg scenario with the existing Jinja function `get_pipeline_id`, which resolves a pipeline Id for a given app/pipeline name.  The use case of this function is to reference Pipeline A from Pipeline B, allowing things like “Run Pipeline B when Pipeline A fails/completes”.  Problem is, when creating a pipeline for the first time `get_pipeline_id` returns None, leading to bugs like Pipeline B never triggering as it doesn't have the Pipeline A Id.

To solve this I’ve introduce the above function `generate_predictable_pipeline_id` which can be used when `get_pipeline_id` fails, like so:

```jinja2
{# Get Pipeline ID if exists #}
{%- set trigger_pipeline_id = get_pipeline_id(app_name, pipeline_name) %}
{%- if not trigger_pipeline_id %}
{# Pipeline does not exist, generate ID #}
{%- set trigger_pipeline_id = generate_predictable_pipeline_id(app_name, pipeline_name) %}
{%- endif %}
{# … later set the Id to the variable: #}
{
  "id": "{{ trigger_pipeline_id }}"
}
```

In other words, the first time Foremast generates this pipeline you can force a certain pipeline ID.  The second+ times you can use an existing Id retrieved from the Spinnaker APIs.

### DIFF
--- a/src/foremast/pipeline/jinja_functions.py
+++ b/src/foremast/pipeline/jinja_functions.py
@@ -16,7 +16,7 @@
 """Functions and variables that can be exposed to Jinja2 templates"""
 
 from ..utils.kayenta import get_canary_id
-from ..utils.pipelines import get_pipeline_id
+from ..utils.pipelines import get_pipeline_id, generate_predictable_pipeline_id
 
 
 def get_jinja_functions():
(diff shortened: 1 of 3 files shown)


linjmeyer:
Docs on UUID5 spec and Python implementation: https://docs.python.org/3/library/uuid.html#uuid.uuid5

PR #443, opened 2020-11-13
feat: Manual templates now support default stage templates

linjmeyer:
This PR allows manual templates to reference Foremast's default Jinja templates and variables used in other pipeline types.

Changes:

* Manual templates now use the same Jinja2 Environment as all other templates
* Manual templates now have access to Jinja2 variables exposed to other templates (e.g. `data.app.pipeline`, `data.app.appname`, etc)
* Existing manual template variables marked as deprecated: `trigger_job`, `group_name`, `app_name`, `repo_name`.  These are now passed in consistently with non-manual templates.  Leaving them in for now for backwards compatibility.

Example:

In a manual template I can now use Foremast's pipeline notification features by doing something like:
```jinja2
{
        "notifications": {% include "pipeline/pipeline_notifications.json.j2" %},
}
```

------------

I noticed the code to pass in variables to JInja templates during pipeline templating is almost identical and in 4-5 places (ec2, manual, lambda, etc.).  That should be consolidated, this is a good first step, but the rest should be a new PR as it will need more testing.

### DIFF
--- a/src/foremast/pipeline/create_pipeline_manual.py
+++ b/src/foremast/pipeline/create_pipeline_manual.py
@@ -97,7 +97,7 @@ class SpinnakerPipelineManual(SpinnakerPipeline):
             # Using lstrip or strip can sometimes remove additional chars
             # We know the string starts with "templates://" already, so remove the exact
             # number of chars to be safe
-            schema_length = len("templatees://")
+            schema_length = len("templates://")
             file_name = file_name[schema_length::]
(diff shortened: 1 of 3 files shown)


jvasallo:
Good clean up of copied code. In addition good work maintaining backwards compatibility. In the future function move to using the standard app block if possible but for now we support both. This is reeling in the manual pipeline type to be more like a standard pipeline type versus a one-off. Nice work

PR #450, opened 2021-01-27
feat: Added support for custom tags on ec2 deploys

jvasallo:
Initial commit getting custom tags to work in deploy stage. Need to continue and add support for other deploy types and custom infra.

- [x] EC2
- [x] ELB
- [x] Lambda
- [x] S3
- [x] etc

### DIFF
--- a/src/foremast/templates/configs/configs.json.j2
+++ b/src/foremast/templates/configs/configs.json.j2
@@ -5,6 +5,7 @@
         "approval_timeout": null,
         "archaius_enabled": false,
         "canary": false,
+        "custom_tags": {},
         "email": null,
         "eureka_enabled": false,
         "instance_profile": "{{ profile }}",
--- a/src/foremast/templates/pipeline/stage-deploy.json.j2
+++ b/src/foremast/templates/pipeline/stage-deploy.json.j2
@@ -35,6 +35,11 @@
           "suspendedProcesses": [],
           "securityGroups": {{ data.app.instance_security_groups }},
           "tags": {
+            {% if data.app.custom_tags %}
+                {% for tag_key, tag_value in data.app.custom_tags.items() -%}
+                  "{{ tag_key }}": "{{ tag_value }}",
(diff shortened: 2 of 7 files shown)


jvasallo:
ELB in Spinnaker currently does not support tagging but can be added later when fixed. 

Once tests pass this should be good.

src/foremast/templates/pipeline/stage-deploy.json.j2
```diff
@@ -35,6 +35,11 @@
           "suspendedProcesses": [],
           "securityGroups": {{ data.app.instance_security_groups }},
           "tags": {
+            {% if data.app.custom_tags %}
```
linjmeyer, line 38: I think tags is missing the closing bracket and comma, and the loop creates a comma on the last item.  Maybe `tags: {{ data.app.custom_tags | tojson }},` would be easier?

src/foremast/templates/configs/configs.json.j2
```diff
@@ -5,6 +5,7 @@
         "approval_timeout": null,
         "archaius_enabled": false,
         "canary": false,
+        "custom_tags": {},
```
linjmeyer, line 8: One way to add tags for all the AWS pipeline types is great 👍

linjmeyer (changes requested):

src/foremast/templates/configs/configs.json.j2
jvasallo: Could extend to GCP too if ever needed. :)

src/foremast/templates/pipeline/stage-deploy.json.j2
jvasallo: @linjmeyer We traditionally always tagged `app_group`, `app_name`, and `owner_email` as you see below. 👍 

We could change it `tojson` however, we might need to refactor to include it in the ec2 code.

PR #460, opened 2021-08-17
Support for Specifying Lambda Subnet Purpose

jvasallo:
Defaulting to legacy "internal" setting, but overriding to allow teams to set specific subnet purpose su


jvasallo:
Will need to update internal templates after merge.

linjmeyer (approved):
lgtm

PR #461, opened 2021-09-08
feat: Ability to add webhooks to pipeline completion

linjmeyer:
Added ability to define webhooks that will fire at the end of each deployment environment.  This is defined in `application-master-$env.json` file:

image


jvasallo:
Nice! Good feature to have. Adds some non-defined webhooks too if teams choose to use em!

PR #465, opened 2021-10-18
feat: Friendly error when env is not defined in pipeline json files

linjmeyer:
Added a check in Foremast runner to verify the correct env keys are present and if not raise a helpful error:

New: `src.foremast.exceptions.ForemastError: Environment 'env_name' not found in pipeline configs.Check pipeline.json and application-master-env_name.json`

Old: `KeyError: 'env_name'`


linjmeyer:
Fixed, saw you approved so went ahead and merged after checks passed

PR #471, opened 2021-11-03
feat: Added app to pipeline_args to pass to Manual Jinja2 templates

jvasallo:
Using App allows to render an app_name that is filtered on organization
This is useful if using gitlab.com for example:

- gitlab.com/org/group/repo: repogroup
- gitlab.com/org/repo: repoorg
- github.com/org/repo: repoorg
- git.selfhosted.com/group/sub/repo: reposub
- git.selfhosted.com/group/repo: repogroup

### DIFF
--- a/src/foremast/pipeline/create_pipeline_manual.py
+++ b/src/foremast/pipeline/create_pipeline_manual.py
@@ -72,7 +72,7 @@ class SpinnakerPipelineManual(SpinnakerPipeline):
 
             # Create all pipelines
             for pipeline_dict in pipelines:
-                pipeline_dict.setdefault('application', self.app_name)
+                pipeline_dict.setdefault('application', self.app)
                 pipeline_dict.setdefault('name',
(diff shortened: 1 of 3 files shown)


src/foremast/pipeline/create_pipeline_manual.py
```diff
@@ -72,7 +72,7 @@ def create_pipeline(self):
 
             # Create all pipelines
             for pipeline_dict in pipelines:
-                pipeline_dict.setdefault('application', self.app_name)
+                pipeline_dict.setdefault('application', self.app)
```
linjmeyer, line 75: This makes it a breaking change right?  so in a template `{{application}}` would need to become something like `{{application.app_name}}` instead?

linjmeyer: We do have that internally on a few manual templates e.g. Spinnaker pipeline json:

```
"application": "{{ application }}",
```

PR #472, opened 2021-11-30
Fix: Added missing arg and updated alias logic to work with update/cr…

dnava13:
- Added missing arg to _put_provisioned_throughput function
- Moved alias check to create/update lambda functions, since it's needed for _put_provisioned_throughput function


linjmeyer (approved):
Nice LGTM.  Saw the linting errors but if this is blocking devs feel free to merge and fix after.

